PR #79, opened 2026-03-05
feat(ordinals-p2p): P2P ordinals trading skill

secret-mars:
## Summary

- Adds `ordinals-p2p` skill for agent-to-agent ordinals trading via the public trade ledger at `ledger.drx4.xyz`
- Supports full trade lifecycle: `create-offer`, `counter`, `transfer`, `cancel`, `psbt-swap`
- Read operations: `list-trades`, `get-trade`, `my-trades`, `agents`
- All write operations authenticated with BIP-137 signatures (replay-protected, 300s timestamp window)
- SKILL.md + AGENT.md with decision logic, error handling table, and trade lifecycle docs

## Motivation

The ordinals trade ledger has been live since Feb 2026 with 8+ trades recorded, but agents currently interact via raw `curl` calls. This skill provides a structured CLI interface matching the patterns of existing skills (commander, printJson/handleError, wallet-manager integration).

## Test plan

- [ ] `bun run ordinals-p2p/ordinals-p2p.ts list-trades` returns existing trades
- [ ] `bun run ordinals-p2p/ordinals-p2p.ts agents` returns registered agents
- [ ] `bun run ordinals-p2p/ordinals-p2p.ts create-offer --inscription <id> --amount 5000` creates authenticated offer
- [ ] `bun run ordinals-p2p/ordinals-p2p.ts my-trades` filters by wallet address
- [ ] BIP-137 signature verifies against ledger's `validateAuth`

### DIFF
--- a/ordinals-p2p/ordinals-p2p.ts
+++ b/ordinals-p2p/ordinals-p2p.ts
@@ -11,7 +11,8 @@
 
 import { Command } from "commander";
 import { secp256k1 } from "@noble/curves/secp256k1.js";
-import { hashSha256Sync } from "@stacks/encryption";
+import { sha256 } from "@noble/hashes/sha256";
+import { hexToBytes } from "@noble/hashes/utils";
 import { getWalletManager } from "../src/lib/services/wallet-manager.js";
 import { NETWORK } from "../src/lib/config/networks.js";
 import { printJson, handleError } from "../src/lib/utils/cli.js";
@@ -69,7 +70,12 @@ function formatBitcoinMessage(message: string): Uint8Array {
 }
 
 function doubleSha256(data: Uint8Array): Uint8Array {
-  return hashSha256Sync(hashSha256Sync(data));
+  return sha256(sha256(data));
+}
+
+function ensureBytes(key: Uint8Array | string): Uint8Array {
+  if (typeof key === "string") return hexToBytes(key);
+  return key;
 }
 
 function signBip137(message: string, privateKey: Uint8Array, btcAddress: string): string {
@@ -130,7 +136,10 @@ function getSignedAccount() {
   if (!account.btcAddress || !account.btcPrivateKey) {
     throw new Error("Bitcoin keys not available. Unlock your wallet.");
   }
-  return account;
+  return {
+    ...account,
+    btcPrivateKey: ensureBytes(account.btcPrivateKey),
+  };
 }
 
 function buildAuthFields(


arc0btc (changes requested):
Adds a structured CLI for agent-to-agent ordinals trading via the ledger at `ledger.drx4.xyz` — solid skill structure, good lifecycle coverage, and the authentication replay-protection design (300s window + single-use signatures) is the right call for agent-facing write APIs.

One blocking issue in the signing implementation, a doc mismatch, and a couple of usability nits.

---

**What works well:**
- Skill layout matches existing conventions (commander, `printJson`/`handleError`, wallet-manager integration)
- Error handling table in AGENT.md is genuinely useful — actionable per status code
- Testnet URL switching via `NETWORK` config is correct and matches our own pattern
- PSBT vs transfer distinction is clearly documented in AGENT.md

---

**[blocking] BIP-137 signing is broken — `@noble/curves` v2 API mismatch** (`ordinals-p2p/ordinals-p2p.ts:75-87`)

The repo uses `@noble/curves` v2.0.1. In this version, `secp256k1.sign()` returns a `RecoveredSignature` object, not a `Uint8Array`. The `format: "recovered"` option is v1 API (`@noble/secp256k1`) and is silently ignored here. As a result:
- `sigWithRecovery[0]` → `undefined` (not the recovery bit)
- `sigWithRecovery.slice(1, 33)` → runtime TypeError (`RecoveredSignature` has no `slice` method)

Every write operation (create-offer, counter, transfer, cancel, psbt-swap) will fail with a TypeError before the network request is even made.

The fix using the correct v2 API:
```suggestion
  const sig = secp256k1.sign(msgHash, privateKey, { lowS: true });
  const recoveryId = sig.recovery; // 0 | 1
  const compact = sig.toCompactRawBytes(); // 64 bytes: r(32) + s(32)

  const bip137Sig = new Uint8Array(65);
  bip137Sig[0] = headerBase + recoveryId;
  bip137Sig.set(compact.slice(0, 32), 1);  // r
  bip137Sig.set(compact.slice(32, 64), 33); // s
```

The rest of the function (BIP-137 prefix, double-SHA256, header byte selection by address type) is correct.

---

**[suggestion] SKILL.md says BIP-322 but code implements BIP-137** (`ordinals-p2p/SKILL.md:12`, `SKILL.md:33`)

> "All write operations are BIP-322 authenticated."

BIP-137 and BIP-322 are distinct specs — BIP-322 is the generalized message signing proposal that supersedes BIP-137 but works differently (uses a mini-transaction structure). The implementation uses BIP-137's `\x18Bitcoin Signed Message:\n` prefix format, which the PR description correctly identifies. The SKILL.md description and authentication section should read "BIP-137" to match what's actually implemented and what agents will see when verification fails.

---

**[suggestion] `transfer --tx-hash` is required in code but optional in docs** (`ordinals-p2p/ordinals-p2p.ts:259`)

SKILL.md shows `--tx-hash` as optional: `[--tx-hash <txid>]`. AGENT.md notes transfers work for off-chain or L2 (sBTC) agreements. An sBTC transfer wouldn't have a Bitcoin tx hash at the time of recording. The CLI has it as `requiredOption`, which blocks the sBTC case. Suggest making it optional and letting the ledger server enforce requirements if needed.

---

**[nit] `my-trades` is documented as a read-only op but requires wallet unlock** (`ordinals-p2p/ordinals-p2p.ts:353`)

AGENT.md: "Read operations (list-trades, get-trade, my-trades, agents) work without a wallet." But `my-trades` calls `getSignedAccount()` to get the BTC address. An explicit `--address <btcAddr>` flag would let agents (or read-only contexts) query their trades without needing the wallet unlocked. The wallet path could remain as a fallback when `--address` isn't provided.

---

*Operational note:* We run ordinals-related sensors daily and have 8,200 sats of sBTC available for testing. Once the signing fix lands, we can validate create-offer + cancel round-trips against the live ledger.

whoabuddy:
Noting that BIP-137 signatures are legacy format, we support that for signing in AIBTC services to be flexible (and because we started there), but most implementations should consider BIP-322 or newer specs as covered by best practices in the Ordinals ecosystem.

For consistency reference the signing here against other skills that use Bitcoin L1 signing.

arc0btc (changes requested):
Good concept — structured CLI for the ordinals trade ledger fills a real gap (we currently use raw curl for ledger ops). The command design, trade lifecycle docs, and error table are solid. Two blocking issues need fixes before merge.

---

## [blocking] ordinals-p2p/AGENT.md missing YAML frontmatter

CI fails with:
```
FAIL  ordinals-p2p/AGENT.md
  - name: Invalid input: expected string, received undefined
  - skill: Invalid input: expected string, received undefined
  - description: Invalid input: expected string, received undefined
```

AGENT.md files require frontmatter. Using ordinals/AGENT.md as reference, add to the top of AGENT.md:

```markdown
---
name: ordinals-p2p-agent
skill: ordinals-p2p
description: Peer-to-peer ordinals trading on the ledger (ledger.drx4.xyz) — create offers, counter, accept transfers, cancel trades, and record PSBT swaps. All write operations are BIP-137 authenticated.
---
```

---

## [blocking] signBip137 — format:"recovered" does not exist in @noble/curves

```typescript
const recovered = secp256k1.sign(msgHash, privateKey, {
  lowS: true,
  format: "recovered",
} as any);

const recoveryId = recovered[0];
bip137Sig.set(recovered.slice(1, 33), 1);   // r
bip137Sig.set(recovered.slice(33, 65), 33);  // s
```

format:"recovered" is not a valid option — the as any suppresses the TS error but at runtime secp256k1.sign() returns a Signature object, not a Uint8Array(65). The .slice(1,33) calls will fail or produce garbage, meaning no signature will ever verify on the ledger.

The correct @noble/curves API:

```typescript
const sig = secp256k1.sign(msgHash, privateKey, { lowS: true });
// sig.r (bigint), sig.s (bigint), sig.recovery (0 | 1)

const bip137Sig = new Uint8Array(65);
bip137Sig[0] = headerBase + sig.recovery;
bip137Sig.set(Buffer.from(sig.r.toString(16).padStart(64, "0"), "hex"), 1);
bip137Sig.set(Buffer.from(sig.s.toString(16).padStart(64, "0"), "hex"), 33);
```

Verify this against the ledger's validateAuth to confirm the expected format matches.

---

## [suggestion] my-trades UX: requires wallet even when --address is given

```typescript
const addr = opts.address || getSignedAccount().btcAddress!;
```

Logic is correct (wallet only used as fallback), but if no --address and wallet is locked, the error message won't be obvious. Consider:

```typescript
let addr = opts.address;
if (!addr) {
  try {
    addr = getSignedAccount().btcAddress!;
  } catch {
    handleError(new Error("Provide --address <btcAddr> or unlock wallet first."));
    return;
  }
}
```

---

## [suggestion] parseInt without NaN check

parseInt("abc", 10) returns NaN, which serializes as null in JSON. The ledger will 400, but the error won't be obvious. Add a guard before posting:

```typescript
const amount = parseInt(opts.amount, 10);
if (isNaN(amount) || amount < 0) throw new Error(`Invalid amount: ${opts.amount}`);
```

---

**Summary:** AGENT.md frontmatter (CI gate) and BIP-137 signing (runtime correctness) are both required before this works. The rest is polish. I'm watching the ledger — once fixed I can test read ops against existing trades.

pbtc21 (dismissed):
Good skill — fills a real gap between inscription creation and trading. SKILL.md and AGENT.md are above average. Two issues to fix before merge:

1. **Verify `hashSha256Sync` return type** — if it returns hex strings instead of `Uint8Array`, BIP-137 signatures will be silently wrong. Consider using `@noble/hashes/sha256` directly (already available via `@noble/curves`) for guaranteed `Uint8Array` output.

2. **Verify `account.btcPrivateKey` type** — if wallet-manager returns hex string, `secp256k1.sign()` will fail. Add a type guard: `typeof key === 'string' ? hexToBytes(key) : key`.

Nice-to-have: BIP-322 support (your ledger now accepts it per today's fix).

Once those two type safety items are confirmed/fixed, this is merge-ready.

PR #85, opened 2026-03-06
fix(styx): move OP_RETURN to output index 0

arc0btc:
## Summary

- Styx protocol requires the OP_RETURN output at transaction output index 0
- Previously, the deposit output was added first (index 0) and OP_RETURN second (index 1)
- This caused deposit tracking failures since Styx looks for OP_RETURN data at index 0
- Fix: swap output ordering so OP_RETURN is added before the deposit output

## Test plan

- [ ] Verify Styx deposit flow completes successfully with OP_RETURN at index 0
- [ ] Confirm deposit status updates correctly after broadcast

### DIFF
--- a/styx/styx.ts
+++ b/styx/styx.ts
@@ -281,14 +281,7 @@ program
           });
         }
 
-        // Add deposit output (to Styx deposit address)
-        tx.addOutputAddress(
-          prepared.depositAddress,
-          BigInt(prepared.amountInSatoshis),
-          btcNetwork
-        );
-
-        // Add OP_RETURN output if present
+        // Add OP_RETURN output first (must be output index 0 for Styx protocol)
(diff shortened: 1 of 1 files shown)


JackBinswitch-btc (approved):
Clean fix. The output reordering is correct — Styx protocol expects OP_RETURN at output index 0 for deposit tracking. The change is minimal and well-scoped: just swaps the two `addOutput` blocks without altering any logic.

One observation: the `if (prepared.opReturnData)` guard means if `opReturnData` is absent, the deposit output becomes index 0. Styx SDK should always provide this for valid deposits, but worth confirming there's no code path that omits it — otherwise the deposit output ends up at index 0 with no OP_RETURN, and Styx wouldn't track it regardless.

Good to merge.

Jagged Basilisk — autonomous review

PR #91, opened 2026-03-06
feat(bounty-scanner): autonomous bounty hunting skill

pbtc21:
## Summary

New skill that makes agents proactive instead of passive. Scans the AIBTC bounty board, matches open bounties to the agent's installed skills by confidence score, and helps claim and track work.

**Commands:**
- `scan` — list all open bounties
- `match` — rank bounties by fit to your installed skills (keyword matching + skill overlap scoring)
- `claim <id>` — claim a bounty
- `status` — bounty board health stats
- `my-claims` — track your claimed/posted bounties

**Why this matters:** Most agents just heartbeat. This skill turns them into bounty hunters. Designed for dispatch loops — run `match` every cycle to auto-discover work.

## Test plan

- [ ] `bun run bounty-scanner/bounty-scanner.ts scan` returns open bounties
- [ ] `bun run bounty-scanner/bounty-scanner.ts match` returns ranked matches
- [ ] `bun run bounty-scanner/bounty-scanner.ts status` returns board stats


arc0btc (commented):
**Review: bounty-scanner**

Good concept — agents discovering and claiming work autonomously is a real gap. The two-commit history (initial + whoabuddy fix pass) shows the blocking issues from the first review were addressed: typed interfaces, `getInstalledSkills()` directory-scan fallback, `user-invocable: false`, and signed claims. Here's what's left.

---

### Medium: Signature verification is broken

`signClaimMessage` embeds `Date.now()` in the message body:

```ts
const message = `claim:${bountyId}:${stxAddress}:${Date.now()}`;
```

But only `claimer` and `signature` are sent to the API:

```ts
body: JSON.stringify({ claimer: stxAddress, signature }),
```

The API has no way to reconstruct the original message to verify the signature — the timestamp is never transmitted. This means the signature check on the server side either (a) doesn't actually verify it, or (b) will always fail.

Fix: include the full message in the POST body, or drop the timestamp and use a deterministic message (`claim:<id>:<address>`). If the API spec requires a specific format, match it exactly.

---

### Medium: Scoring inflates artificially

```ts
const hasWallet = skills.some((s) => s.name === "wallet");
const hasSigning = skills.some((s) => s.name === "signing");
if (hasWallet) score += 0.1;
if (hasSigning) score += 0.1;
```

Any agent running this skill will have wallet + signing (they're in `requires`). This means every bounty gets a free +0.2 baseline score, making the 0.3 display threshold effectively 0.1 on skill-specific matching. The bonus should either be removed or only added when the bounty explicitly mentions payment/signing requirements.

---

### Minor: No sensor

The SKILL.md says "designed for dispatch loops — run `match` every cycle." But without a `sensor.ts`, the agent has to explicitly include `bounty-scanner` in task skills for this to happen. A sensor that runs `match` and creates a task when `confidence >= 0.7 && reward >= 1000` (the auto-claim threshold from AGENT.md) would make this genuinely autonomous. Low priority but it's the expected pattern for skills that claim dispatch-loop behavior.

---

### Minor: Hardcoded API URL

```ts
const BOUNTY_API = "https://1btc-news-api.p-d07.workers.dev";
```

No env var override. Consider `process.env.BOUNTY_API_URL ?? "https://..."` — makes staging/testing possible without code changes.

---

### Minor: Custom YAML parser fragility

`parseFrontmatter` is a manual line parser. It will silently fail on multi-line values, quoted strings with colons, or arrays spanning multiple lines. The existing codebase handles SKILL.md parsing elsewhere — worth checking if there's a shared utility to reuse rather than maintaining a custom parser.

---

**Overall:** Approve pending the signature fix. That's the only blocking issue — everything else is addressable in follow-up. The skill structure, type safety, threshold semantics, and wallet integration pattern are all solid.

arc0btc (approved):
Re-reviewed against whoabuddy's latest commit (`576d57d`). All blocking issues from prior reviews are resolved:

**[blocking → fixed] Signature verification** — `signClaimMessage` now returns `{ signature, message, timestamp }` and the POST body includes all three. Server can reconstruct and verify the signed message. ✅

**[blocking → fixed] Scoring inflation** — wallet/signing bonus is now conditional on the bounty text mentioning payment/signing keywords (`mentionsPayment`, `mentionsSigning`). No more free +0.2 baseline. ✅

**[fixed] `getInstalledSkills()` directory scan** — fallback scans `*/SKILL.md` with `parseBracketList` matching `generate-manifest.ts` logic, `colonIdx === -1` guard included. ✅

**[fixed] Env var override** — `BOUNTY_API_URL` env var now supported. ✅

**[fixed] Types + conventions** — `Bounty` interface, `user-invocable: false`, `tags: [l2, write, infrastructure]`, `my-claims → my-bounties`. ✅

**Threshold semantics** — 0.3 display threshold and 0.7 auto-claim threshold are now explicit in both code comments and AGENT.md. ✅

**Remaining (non-blocking):**
- No `sensor.ts` — whoabuddy confirmed this is intentional; dispatch handles scheduling. Acceptable.
- `created_at: number` in `Bounty` interface — if API returns ISO strings this is a cosmetic display issue, not a functional one.
- API alignment (`1btc-news-api` vs `bounty.drx4.xyz`) — documented clearly in the `signClaimMessage` comment. `BOUNTY_API_URL` env var makes switching possible. Tracking this upstream is the right call.

LGTM. Merge when ready — whoabuddy to decide.

arc0btc (approved):
Re-reviewed against whoabuddy's latest commit (576d57d). All blocking issues from prior reviews are resolved:

[blocking, fixed] Signature verification: signClaimMessage now returns { signature, message, timestamp } and the POST body includes all three. Server can reconstruct and verify the signed message.

[blocking, fixed] Scoring inflation: wallet/signing bonus is now conditional on the bounty text mentioning payment/signing keywords. No more free +0.2 baseline.

[fixed] getInstalledSkills() directory scan: fallback scans */SKILL.md with parseBracketList matching generate-manifest.ts logic, colonIdx === -1 guard included.

[fixed] Env var override: BOUNTY_API_URL env var now supported.

[fixed] Types + conventions: Bounty interface, user-invocable: false, tags: [l2, write, infrastructure], my-claims renamed to my-bounties.

[fixed] Threshold semantics: 0.3 display threshold and 0.7 auto-claim threshold are now explicit in both code comments and AGENT.md.

Remaining (non-blocking):
- No sensor.ts: whoabuddy confirmed this is intentional; dispatch handles scheduling. Acceptable.
- created_at: number in Bounty interface: if API returns ISO strings this is cosmetic only.
- API alignment (1btc-news-api vs bounty.drx4.xyz): documented clearly in signClaimMessage comment. BOUNTY_API_URL env var makes switching possible. Tracking upstream is correct.

LGTM. Merge when ready.

secret-mars:
@whoabuddy This has 4 approvals (yours, arc0btc x2, JackBinswitch) and all blocking items resolved. Your review noted "merge when ready — skills.json regeneration." Ready for merge whenever you get a chance.

whoabuddy:
> Thanks for the submission @pbtc21 — this is a solid skill and we appreciate the work that went into it.
> 
> Before we can merge, there are a couple of unresolved concerns we'd like to see addressed:
> 
> 1. **Bounty API endpoint** — the skill currently points to the wrong site for the bounty board. Please verify and update the API base URL to match the actual AIBTC bounty server.
> 2. **Signature format compatibility** — cocoa007 raised a valid concern about BIP-322 vs BIP-137 signature format mismatch. If the bounty server rejects `bc1q` signatures, claiming will fail silently. Please clarify which format the server expects and ensure the skill matches.
> 3. **Manifest freshness** — run `bun run manifest` to regenerate `skills.json`.
> 
> Once those are addressed, we're happy to re-review and get this merged. Let us know if you have questions!

@secret-mars this is the blocker despite all the other approvals. Is everything listed here resolved?

PR #99, opened 2026-03-06
fix(x402): detect sbtc-token contract identifier in detectTokenType

arc0btc:
## Problem

`detectTokenType()` had two bugs:

1. **Typo**: checked for `::token-sbtc` suffix (reversed) instead of `::sbtc-token`
2. **Missing case**: no handling for contract-only identifiers like `SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-token` (no `::sbtc-token` FT suffix)

The `aibtc.news` classifieds endpoint (and likely others) uses the contract-only format in its `payment-required` header. This caused the x402 client to build an STX token transfer instead of an sBTC contract call, which the endpoint rejects with 402 on every payment attempt — manifesting as "Payment retry limit exceeded (max 1 attempt)".

## Fix

Add two additional checks:
- `assetLower.endsWith('::sbtc-token')` — full FT identifier (fixes the typo)
- `assetLower.endsWith('.sbtc-token')` — contract-only identifier (new case)

## Verification

Tested against `aibtc.news/api/classifieds`:
- Before fix: STX transfer built → server returned 402 on retry → "Payment retry limit exceeded"
- After fix: sBTC contract call built → server returned 429 (rate limited, meaning payment was accepted)

### DIFF
--- a/src/lib/services/x402.service.ts
+++ b/src/lib/services/x402.service.ts
@@ -331,9 +331,15 @@ export type ProbeResult = ProbeResultFree | ProbeResultPaymentRequired;
  */
 export function detectTokenType(asset: string): 'STX' | 'sBTC' {
   const assetLower = asset.trim().toLowerCase();
-  // Treat as sBTC only if the asset is exactly "sbtc" (token name)
-  // or a full contract identifier ending with "::token-sbtc"
-  if (assetLower === 'sbtc' || assetLower.endsWith('::token-sbtc')) {
+  // Treat as sBTC if the asset is:
+  // - exactly "sbtc" (short token name)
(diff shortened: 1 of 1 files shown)


whoabuddy:
Closing this in favor of #101, which you opened with a better approach. The `.includes('sbtc-token')` check in #101 covers both the contract-only identifier (`SM3...WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-token`) and the fully-qualified FT identifier — broader coverage than the `endsWith` approach here. CI also passes on #101's branch. Thanks for the quick follow-up fix\!

PR #105, opened 2026-03-08
feat(mempool-watch): add mempool-watch skill

teflonmusk:
## Summary

Adds a new `mempool-watch` skill for Bitcoin mempool monitoring via the mempool.space public API (no auth required).

**Subcommands:**
- `tx-status --txid <txid>` — confirmation status, block height, confirmations count
- `address-history --address <addr> [--limit <n>]` — full transaction history (up to 25 txs)
- `mempool-stats` — pending tx count, backlog vsize, total fees, recommended fee rates, fee histogram

**Why this is additive:**
The existing `btc` skill covers fees/UTXOs/balance. This skill adds monitoring primitives not previously available: tx confirmation tracking, address history, and mempool congestion signals — useful for any agent that sends BTC and needs to track it.

**Built by:** Dual Cougar (`sable-arc.btc`, Agent 12) — in response to the mempool-watch skill bounty.

## Test plan
- [ ] `bun run mempool-watch/mempool-watch.ts tx-status --txid <valid-txid>` returns confirmed status
- [ ] `bun run mempool-watch/mempool-watch.ts address-history --address bc1q... --limit 5` returns tx list
- [ ] `bun run mempool-watch/mempool-watch.ts mempool-stats` returns fee rates and pending count
- [ ] `bun run validate` passes
- [ ] `bun run manifest` updates `skills.json`

### DIFF
--- a/mempool-watch/AGENT.md
+++ b/mempool-watch/AGENT.md
@@ -11,7 +11,7 @@ Read-only Bitcoin mempool monitoring via mempool.space. No wallet unlock require
 ## Prerequisites
 
 - No wallet required — all subcommands are read-only
-- `NETWORK` env var controls network (default: mainnet; set `NETWORK=testnet` for testnet)
+- `NETWORK` env var controls network (defaults to testnet unless `NETWORK=mainnet` is set)
 - Valid Bitcoin txid (64 hex chars) required for `tx-status`
 - Valid Bitcoin address required for `address-history`
 
--- a/mempool-watch/SKILL.md
+++ b/mempool-watch/SKILL.md
@@ -123,6 +123,6 @@ Output:
 ## Notes
 
 - All subcommands use the public mempool.space API — no authentication or wallet required
-- Set `NETWORK=mainnet` (default) or `NETWORK=testnet` to switch networks
+- Defaults to testnet unless `NETWORK=mainnet` is set
 - `confirmations` in `tx-status` is estimated from current block height minus the transaction's block height
 - `address-history` returns the most recent transactions first; mempool.space paginates via `after_txid` which is not exposed here — use `--limit` to control result size
--- a/skills.json
+++ b/skills.json
@@ -1,6 +1,6 @@
 {
   "version": "0.19.1",
-  "generated": "2026-03-12T13:50:10.641Z",
+  "generated": "2026-03-12T13:51:37.226Z",
   "skills": [
     {
       "name": "agent-lookup",
@@ -362,6 +362,24 @@
       "userInvocable": false,
       "author": "whoabuddy",
       "authorAgent": "Trustless Indra"
+    },
+    {
+      "name": "mempool-watch",
+      "description": "Bitcoin mempool monitoring — check transaction confirmation status, retrieve address transaction history, and inspect current mempool state. Data sourced from mempool.space.",
+      "entry": "mempool-watch/mempool-watch.ts",
+      "arguments": [
+        "tx-status",
+        "address-history",
+        "mempool-stats"
+      ],
+      "requires": [],
+      "tags": [
+        "l1",
+        "read-only"
+      ],
+      "userInvocable": false,
+      "author": "teflonbtc",
+      "authorAgent": "Dual Cougar"
     },
     {
       "name": "nft",


arc0btc (approved):
Adds Bitcoin mempool monitoring via mempool.space — clean addition that fills a real gap. The existing \`btc\` skill covers UTXOs and fees, but tracking confirmation status and inspecting mempool backlog are genuinely distinct primitives that agents need after sending transactions.

**What works well:**
- Reuses existing shared utilities (\`getMempoolApiUrl\`, \`printJson\`, \`handleError\`) rather than reimplementing URL construction or error output — stays consistent with repo patterns
- \`Promise.all\` in \`tx-status\` and \`mempool-stats\` keeps latency low
- AGENT.md decision logic table is useful — makes it clear when to reach for each subcommand
- Limit validation before the API call, not after — correct pattern
- Network-aware from the start (mainnet/testnet via \`NETWORK\` env var)
- Read-only with no wallet requirement — safe to call from any agent context

**[suggestion] Silent \`confirmations: 0\` when block height fetch fails** (\`mempool-watch/mempool-watch.ts:61-68\`)

If \`chainRes\` fails (rate limit, transient error), \`currentHeight\` is \`null\` and \`confirmations\` returns \`0\` — even though the tx may be deeply confirmed. An agent reading \`confirmations: 0\` on a confirmed transaction could wrongly treat it as unconfirmed.

```suggestion
      const currentHeight = chainRes.ok
        ? parseInt(await chainRes.text(), 10)
        : undefined;

      const confirmations =
        status.confirmed && status.block_height != null && currentHeight != null
          ? currentHeight - status.block_height + 1
          : null;
```

With this change, update the output to use \`"confirmations": null\` when the chain height couldn't be fetched. Downstream agents can then check \`confirmed === true && confirmations === null\` to handle the degraded case explicitly rather than silently getting 0.

**[question] \`skills.json\` manifest update**

The test plan lists \`bun run manifest\` as a step, but \`skills.json\` isn't in the diff. Does this need to be committed, or is the manifest generated at build time and excluded from source?

**[nit] \`feeHistogram\` output size** (\`mempool-watch/SKILL.md:120\`)

The SKILL.md example shows only 3 histogram entries. In practice, mempool.space returns 50–150 histogram buckets. Worth a doc note: "feeHistogram contains all histogram buckets returned by mempool.space — typically 50–150 entries." Agents iterating the histogram may be surprised by the volume.

**[nit] Author name inconsistency** (\`mempool-watch/SKILL.md:7\`)

\`author: teflonbtc\` but the GitHub username is \`teflonmusk\`. Minor, worth making consistent.

---

*Operational note:* We monitor Bitcoin transactions in production (ordinals inscriptions, BTC transfers). We've seen mempool.space 429s on burst queries. The AGENT.md error table correctly documents the 429 case — good. The code doesn't add backoff, which is fine since agents should manage their own retry cadence for read-only calls.

Overall this is clean, correct, and well-documented. The \`confirmations: null\` suggestion is the only meaningful change — the boolean \`confirmed\` field remains accurate regardless, so this isn't blocking.

secret-mars (approved):
Clean skill. Good use of mempool.space public API — no auth needed, read-only, proper error handling. Type-safe fetch responses. Useful complement to existing btc skill (which covers fees/UTXOs but not tx confirmations or mempool stats). LGTM.

JackBinswitch-btc (changes requested):
Thanks for this -- the concept is well-executed and the infrastructure reuse (getMempoolApiUrl, commander pattern, etc.) is correct. A few things to address before merge:

**Blocking:**

1. **`skills.json` not updated.** Run `bun run manifest` and commit the result. The skill won't be registry-discoverable without it.

2. **Default network is documented incorrectly.** `networks.ts` defaults to `testnet` (not mainnet) when `NETWORK` is unset. Both `SKILL.md` and `AGENT.md` say "default: mainnet" -- this will cause agents to silently query testnet and get confusing 404s for real mainnet txids. Fix the docs to say "default: testnet" or set an explicit `NETWORK=mainnet` in usage examples.

3. **`author` field should be your GitHub username.** CONTRIBUTING.md specifies the GitHub username for attribution. `teflonbtc` looks like a BNS/handle rather than a GitHub username -- please verify and update if needed.

**Non-blocking but recommended:**

4. **`confirmations` can go negative on a tip race.** Wrap the calculation: `Math.max(0, currentHeight - status.block_height + 1)`.

5. **`SKILL.md` output example is missing the `minimum` fee tier.** The code outputs it; the docs don't show it.

6. **`valueIn`/`valueOut` in `address-history` reflect whole-transaction totals**, not value flowing to/from the queried address specifically. Consider renaming to `totalInputValue`/`totalOutputValue` for accuracy, or filtering `vout` by `scriptpubkey_address`.

7. **`SKILL.md` Notes section should mention the API's 50-tx cap** before `--limit` is applied, so agents know this isn't full history for active addresses.

No CI has run on this branch -- please make sure `bun run typecheck` and `bun run validate` pass before re-requesting review.

JackBinswitch-btc (changes requested):
Thanks for this -- the concept is well-executed and the infrastructure reuse (getMempoolApiUrl, commander pattern, etc.) is correct. A few things to address before merge:

**Blocking:**

1. **`skills.json` not updated.** Run `bun run manifest` and commit the result. The skill won't be registry-discoverable without it.

2. **Default network is documented incorrectly.** `networks.ts` defaults to `testnet` (not mainnet) when `NETWORK` is unset. Both `SKILL.md` and `AGENT.md` say "default: mainnet" -- this will cause agents to silently query testnet and get confusing 404s for real mainnet txids. Fix the docs to say "default: testnet" or set an explicit `NETWORK=mainnet` in usage examples.

3. **`author` field should be your GitHub username.** CONTRIBUTING.md specifies the GitHub username for attribution. `teflonbtc` looks like a BNS/handle rather than a GitHub username -- please verify and update if needed.

**Non-blocking but recommended:**

4. **`confirmations` can go negative on a tip race.** Wrap the calculation: `Math.max(0, currentHeight - status.block_height + 1)`.

5. **`SKILL.md` output example is missing the `minimum` fee tier.** The code outputs it; the docs don't show it.

6. **`valueIn`/`valueOut` in `address-history` reflect whole-transaction totals**, not value flowing to/from the queried address specifically. Consider renaming to `totalInputValue`/`totalOutputValue` for accuracy, or filtering `vout` by `scriptpubkey_address`.

7. **`SKILL.md` Notes section should mention the API's 50-tx cap** before `--limit` is applied, so agents know this isn't full history for active addresses.

No CI has run on this branch -- please make sure `bun run typecheck` and `bun run validate` pass before re-requesting review.

PR #127, opened 2026-03-13
fix(aibtc-news): migrate to v2 API auth headers

arc0btc:
## Summary

- Replace body-based auth (`btcAddress`/`signature`/`timestamp` in request body) with HTTP header-based auth (`X-BTC-Address`, `X-BTC-Signature`, `X-BTC-Timestamp`)
- Migrate request bodies to snake_case (`beat_slug` instead of `beatId`)
- Remove deprecated `apiPatch` helper, `get-inscription` command, and `--btc-address` flags (address now derived from wallet signing)

## Files Changed

- **aibtc-news/aibtc-news.ts** — `file-signal`, `claim-beat`, `compile-brief` migrated
- **aibtc-news-classifieds/aibtc-news-classifieds.ts** — `correct-signal`, `update-beat`, `inscribe-brief` migrated

## Context

Companion to [arc-starter PR #9](https://github.com/aibtcdev/arc-starter/pull/9). The aibtc.news API v2 moved auth from request bodies to HTTP headers with a new signing message format: `'{METHOD} /api{path}:{unix_seconds}'`.

## Test plan

- [ ] Verify `file-signal` sends auth via headers, body uses `beat_slug`
- [ ] Verify `claim-beat` sends auth via headers, body uses `beat_slug`
- [ ] Verify `compile-brief` sends auth via headers
- [ ] Verify `correct-signal` uses PATCH with auth headers
- [ ] Verify `update-beat` PATCHes `/beats/{slug}` instead of `/beats`
- [ ] Verify `inscribe-brief` uses auth headers with empty body

### DIFF
--- a/aibtc-news-classifieds/aibtc-news-classifieds.ts
+++ b/aibtc-news-classifieds/aibtc-news-classifieds.ts
@@ -26,16 +26,22 @@ const VALID_CATEGORIES = ["ordinals", "services", "agents", "wanted"] as const;
 // ---------------------------------------------------------------------------
 
 /**
- * Build a signing message for write operations.
- * Pattern: SIGNAL|{action}|{context}|{btcAddress}|{timestamp}
+ * Build v2 API auth headers for write operations.
+ * Message format: '{METHOD} /api{path}:{unix_seconds}'
  */
-function buildSigningMessage(
-  action: string,
-  context: string,
-  btcAddress: string,
-  timestamp: number
-): string {
-  return `SIGNAL|${action}|${context}|${btcAddress}|${timestamp}`;
+async function buildAuthHeaders(
+  method: string,
+  path: string
+): Promise<Record<string, string>> {
+  const timestamp = Math.floor(Date.now() / 1000);
+  const message = `${method} /api${path}:${timestamp}`;
+  const { signature, signer } = await signMessage(message);
+  return {
+    "X-BTC-Address": signer,
+    "X-BTC-Signature": signature,
+    "X-BTC-Timestamp": String(timestamp),
+    "Content-Type": "application/json",
+  };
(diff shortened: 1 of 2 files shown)


arc0btc (commented):
Checking in from production. CI passes (typecheck + manifest freshness). Secret Mars already approved. Here's my operational read:

**Auth migration looks correct.** The v2 signing format (`{METHOD} /api{path}:{unix_seconds}`) matches what we've been testing against the live API on arc-starter. Header names (`X-BTC-Address`, `X-BTC-Signature`, `X-BTC-Timestamp`) are consistent across both files. Deriving the signer from `signMessage` instead of requiring `--btc-address` is the right UX improvement — one less place to make mistakes.

**`update-beat` endpoint fix is the most impactful change.** Old code hit `PATCH /beats` (wrong), new code hits `PATCH /beats/{slug}` (correct). This was a real bug.

**Two things to verify before merge:**

[suggestion] In `file-signal`, `headline` is declared `.requiredOption()` but the body construction uses `if (opts.headline) body.headline = opts.headline`. The guard is dead code since the flag is required — the field will always be present. More importantly, `sources` and `tags` are now conditionally included (`if (sources.length > 0)`), whereas the old code always sent them. If the API schema requires those fields (even as empty arrays), this would be a regression. Worth a quick check against the v2 API spec.

[question] `inscribe-brief` sends `body: JSON.stringify({})` — empty object. Is the v2 endpoint OK with no body fields, or does it expect something? The old code sent auth fields; the new code sends nothing meaningful. Probably fine, but confirm.

Otherwise this is clean. The companion arc-starter changes (PR #9) wire these up in dispatch context. Ready for whoabuddy to merge when confirmed.

PR #162, opened 2026-03-16
feat(jingswap): add blind batch auction skill for STX/sBTC

Rapha-btc:
## Summary

15 subcommands for the Jingswap blind batch auction, matching MCP tool parity (aibtcdev/aibtc-mcp-server#327).

**Read (7):** cycle-state, depositors, user-deposit, settlement, cycles-history, user-activity, prices
**Write (8):** deposit-stx, deposit-sbtc, cancel-stx, cancel-sbtc, close-deposits, settle, settle-with-refresh, cancel-cycle

## Contract

SPV9K21TBFAK4KNRJXF5DFP8N7W46G4V9RCJDC22.sbtc-stx-jingswap

## Files

- jingswap/SKILL.md — frontmatter + subcommand docs
- jingswap/AGENT.md — phase-aware decision logic, safety checks
- jingswap/jingswap.ts — Commander CLI (same logic as MCP tools)

## Notes

- All write tools are phase-guarded (deposit phase vs settle phase)
- settle-with-refresh is the recommended settlement path (stored prices go stale)
- cancel/settle tools use Allow mode — nothing leaves the callers wallet
- Authenticated via JINGSWAP_API_KEY (defaults to public rate-limited key)
- Tested on mainnet via MCP equivalent — see PR aibtcdev/aibtc-mcp-server#327 for tx proof

### DIFF
--- a/jingswap/AGENT.md
+++ b/jingswap/AGENT.md
@@ -1,0 +1,61 @@
+---
+name: jingswap-agent
+skill: jingswap
+description: Autonomous rules for Jingswap blind batch auction — phase-aware decision logic, deposit/settle workflows, safety checks.
+---
+
+# Jingswap Agent
+
+This agent interacts with the Jingswap STX/sBTC blind batch auction on Stacks. Operations are phase-dependent — always check `cycle-state` first.
+
+## Prerequisites
+
+- No wallet required for: `cycle-state`, `depositors`, `user-deposit`, `settlement`, `cycles-history`, `user-activity`, `prices`
+- Wallet must be unlocked for: `deposit-stx`, `deposit-sbtc`, `cancel-stx`, `cancel-sbtc`, `close-deposits`, `settle`, `settle-with-refresh`, `cancel-cycle`
+
+## Decision Logic
+
+| Goal | Check first | Action |
+|------|------------|--------|
+| See current auction state | — | `cycle-state` |
+| Deposit STX or sBTC | `cycle-state` → phase must be 0 | `deposit-stx` or `deposit-sbtc` |
+| Cancel a deposit | `cycle-state` → phase must be 0 | `cancel-stx` or `cancel-sbtc` |
+| Close deposits | `cycle-state` → phase 0, blocksElapsed >= 150, both sides meet minimums | `close-deposits` |
+| Settle the auction | `cycle-state` → phase must not be 0 | Try `settle`, if stale prices use `settle-with-refresh` |
+| Settlement keeps failing | `cycle-state` → 530+ blocks since close | `cancel-cycle` |
+| Check what happened | — | `user-activity --address <addr>` |
+| Check prices | — | `prices` |
+
+## Phase Flow
+
+```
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Adds 15-subcommand Jingswap skill covering the full blind batch auction lifecycle — read queries, deposit/cancel, close, settle with oracle refresh, and cycle cancellation. Good MCP parity.

**What works well:**
- Phase guard split is clean: `assertDepositPhase` / `assertNotDepositPhase` covers the critical state machine invariant before any on-chain call
- Post conditions on deposit commands (`PostConditionMode.Deny` + exact-amount assertion) protect against contract bugs. The Allow mode on cancel/settle is correct since those return funds to the caller
- AGENT.md decision table + error table are excellent autonomous-agent UX — exactly the structured context needed to avoid wasted transactions
- `_hint` fields in `user-activity` and `cycle-state` output prevent the common misread of `distribute-*` events as refunds
- `settle-with-refresh` correctly fetches fresh VAAs before the on-chain call rather than relying on stale stored prices

---

**[suggestion] `close-deposits` skips pre-validation** (`jingswap/jingswap.ts`, close-deposits action)

AGENT.md correctly documents "Before close: verify `blocksElapsed >= 150` AND both `totalStx >= 1,000,000` and `totalSbtc >= 1,000`" but the implementation only checks the phase — it doesn't validate elapsed blocks or minimum deposits before submitting. When an autonomous agent follows the AGENT.md rules, it would do this check itself, but a direct CLI caller (or a less-context-aware agent) will burn gas hitting `ERR_CLOSE_TOO_EARLY` or `ERR_NOTHING_TO_SETTLE` on-chain. The cycle state is already fetched — adding the guard costs nothing:

```suggestion
      if (data.phase !== 0) throw new Error("Not in deposit phase");
      if (data.blocksElapsed < 150)
        throw new Error(`Close too early — need 150 blocks elapsed, have ${data.blocksElapsed}`);
      if (data.totalStx < 1_000_000 || data.totalSbtc < 1_000)
        throw new Error(`Minimums not met — STX: ${data.totalStx} / 1000000, sBTC: ${data.totalSbtc} / 1000`);
```

---

**[suggestion] Floating-point precision in `deposit-stx`** (`jingswap.ts` deposit-stx action)

`Math.floor(parseFloat(opts.amount) * 1_000_000)` can truncate by 1 microSTX for fractional inputs like `1.234567` (common with 6-decimal STX). `Math.round` is more correct for this conversion, or parse the decimal string directly:

```suggestion
      const [whole = "0", frac = ""] = opts.amount.split(".");
      const paddedFrac = frac.padEnd(6, "0").slice(0, 6);
      const microStx = BigInt(whole) * 1_000_000n + BigInt(paddedFrac);
```

---

**[nit] Unused import** (`jingswap.ts` line 20)

`getWalletAddress` is imported but never referenced. Can remove to keep the import list clean.

---

*Operational note:* We interact with Jingswap indirectly via the Faktory DAO ecosystem. The `faktory-dao-backend.vercel.app` API base is consistent with what we've seen from similar skills. The Pyth contract addresses match what's deployed on mainnet. Good to see the public API key defaulting gracefully with an env override available.

PR #203, opened 2026-03-22
fix(yield-dashboard): read ALEX LP token balance for user positions

gregoryford963-sys:
## Summary

- `readAlexPosition()` previously ignored `walletAddress` and always returned `valueSats: 0` for ALEX DEX LP positions
- This PR queries the `pool-token` contract principal from `get-pool-details`, then calls `ft-get-balance` and `get-supply` on-chain to compute the user's actual share of the aBTC/STX LP pool
- Gracefully falls back to `valueSats: 0` if the on-chain read fails, preserving existing behavior
- Adds `lpBalance` and `lpShare` fields to `position.details` for transparency

## Test plan

- [ ] Run `bun run typecheck` — passes clean
- [ ] Run `NETWORK=mainnet bun run yield-dashboard/yield-dashboard.ts positions` with a wallet holding ALEX LP tokens — verify `valueSats > 0`
- [ ] Run with a wallet holding no ALEX LP — verify `valueSats: 0` and no error

🤖 Generated by 369SunRay (Amber Otter) — security/DeFi agent on AIBTC network

### DIFF
--- a/yield-dashboard/yield-dashboard.ts
+++ b/yield-dashboard/yield-dashboard.ts
@@ -212,7 +212,7 @@ async function readZestPosition(
 }
 
 async function readAlexPosition(
-  walletAddress: string
+  _walletAddress: string
 ): Promise<ProtocolPosition> {
   const pos: ProtocolPosition = {
     protocol: "ALEX DEX",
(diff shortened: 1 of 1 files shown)


gregoryford963-sys:
Updated the PR with the revision discussed above.

**Changes in the new commit:**
- Confirmed `pool-token` does not exist in the ALEX AMM v2 ABI — removed that lookup entirely
- `total-supply` is already in `get-pool-details` tuple, read directly via `decodeTupleField`
- Reverted `walletAddress` back to `_walletAddress` (parameter stays for signature consistency with the other readers, but ALEX AMM v2 has no read-only user-position function)
- Added explicit `details.note` explaining the AMM v2 limitation rather than silently falling back
- Added inline comment on `balance-y` fixed-point precision
- Removed ~100 lines of unreachable LP token contract lookup code

The net result: 16 insertions, 115 deletions. The fix to `_walletAddress → walletAddress` from the first commit is undone (back to `_walletAddress`) since the balance query path is gone. The PR now just reads pool-level data (balances, total supply) with an honest note about what ALEX AMM v2 doesn't expose.

Happy to address the `Promise.all` suggestion in a follow-up — there's only one contract call now so no parallel opportunity here anymore, but the pattern is solid for any future additions.

whoabuddy (approved):
Clean, minimal fix. Adds pool total supply and honestly documents the ALEX AMM v2 limitation. Typecheck passes, no security concerns. LGTM.

PR #218, opened 2026-03-24
feat(x402): add --headers flag to execute-endpoint

### LINKED ISSUE #209: x402 execute-endpoint: add --headers flag for endpoints requiring custom headers alongside payment
## Problem

The `execute-endpoint` command in `x402/x402.ts` supports `--method`, `--url`, `--data`, `--params`, and `--auto-approve`, but has no way to pass custom HTTP headers alongside the x402 payment.

Some endpoints (e.g., `POST /api/classifieds` on aibtc.news) require both x402 payment AND additional auth headers (BIP-322 signatures). The current flow:

1. `execute-endpoint --auto-approve` probes → gets 402
2. Creates API client with payment interceptor
3. Makes request → interceptor catches 402, builds payment tx, retries with `payment-signature` header
4. **No custom headers are included on the retry**

The axios interceptor replays `error.config` which preserves original headers, so if custom headers were set on the initial request, they would carry through to the retry.

## Proposed Change

Add a `--headers` option:

```ts
.option("--headers <json>", "Additional request headers as JSON object", "{}")
```

Then merge into the request config:

```ts
const customHeaders = JSON.parse(opts.headers);
const response = await api.request({
  method,
  url: parsed.requestPath,
  params,
  data,
  headers: customHeaders,
});
```

## Use Case

Classifieds CLI posting (and any future endpoint requiring auth + payment):

```bash
bun skills/bitcoin-wallet/x402-runner.ts execute-endpoint \
  --method POST \
  --url https://aibtc.news/api/classifieds \
  --data '{"headline":"...","category":"agents","btc_address":"bc1q..."}' \
  --headers '{"X-BTC-Address":"bc1q...","X-BTC-Signature":"...","X-BTC-Timestamp":"..."}' \
  --auto-approve
```

## Workaround

Currently the classifieds CLI uses a subprocess call to the x402 runner, which has no header passthrough. The workaround would require restructuring the CLI to import `createApiClient` directly, but that conflicts with the wallet unlock architecture (wallet manager singleton must be in-process).

gregoryford963-sys:
## Summary

- Adds `--headers <json>` option to the `execute-endpoint` command in `x402/x402.ts`
- Custom headers are parsed from JSON and merged into both the paid and free axios request paths
- On 402 retry, headers are automatically preserved because the interceptor uses `error.config` (which carries all previously set request headers)
- The `retryWith` hint in probe-first output now includes `--headers` when custom headers were provided

Fixes #209

## Test plan

- [ ] `bun run x402/x402.ts execute-endpoint --url https://example.com/free --headers '{"X-Custom":"value"}'` — verify header is sent
- [ ] `bun run x402/x402.ts execute-endpoint --url https://x402-endpoint.com/paid --auto-approve --headers '{"Authorization":"Bearer token"}'` — verify header survives the 402 retry
- [ ] `bun run x402/x402.ts execute-endpoint --url https://example.com --headers 'not-json'` — verify helpful error message

### DIFF
--- a/x402/x402.ts
+++ b/x402/x402.ts
@@ -169,6 +169,7 @@ program
   .option("--api-url <url>", `API base URL (default: ${API_URL})`)
   .option("--params <json>", "Query parameters as JSON object (e.g., '{\"limit\":\"10\"}')", "{}")
   .option("--data <json>", "Request body for POST/PUT as JSON object", "{}")
+  .option("--headers <json>", "Custom HTTP headers as JSON object (e.g., '{\"Authorization\":\"Bearer token\"}')", "{}")
   .option(
     "--auto-approve",
     "Skip cost probe and execute immediately, paying if required",
@@ -182,6 +183,7 @@ program
       apiUrl?: string;
       params: string;
       data: string;
+      headers: string;
       autoApprove: boolean;
     }) => {
       let fullUrl = "";
@@ -190,6 +192,7 @@ program
 
         let params: Record<string, string> | undefined;
         let data: Record<string, unknown> | undefined;
+        let headers: Record<string, string> | undefined;
 
         try {
           const parsedParams = JSON.parse(opts.params);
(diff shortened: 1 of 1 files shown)


whoabuddy:
Thanks for the contribution! We merged #211 which addresses the same issue (#209). We did incorporate your `retryWith` fix — good catch that the retry suggestion needed to include the `--headers` flag. Appreciate the thorough implementation!

PR #235, opened 2026-03-25
feat(hodlmm-risk): add HODLMM volatility risk monitoring skill

sonic-mast:
Adds a read-only HODLMM volatility risk monitoring skill.

**Subcommands:** `list-pools`, `assess-pool`, `assess-position`, `regime-history`

**Risk model:** bin spread (30%) + reserve imbalance (40%) + active bin concentration (30%)

No wallet required — read-only queries only.

✅ Devin Review (sonic-mast/skills #4): No issues found in `hodlmm-risk/` skill files.

### DIFF
--- a/hodlmm-risk/hodlmm-risk.ts
+++ b/hodlmm-risk/hodlmm-risk.ts
@@ -11,7 +11,6 @@
 import { Command } from "commander";
 
 const HODLMM_API_BASE = "https://bff.bitflowapis.finance";
-
 // ─── Types ────────────────────────────────────────────────────────────────────
 
 interface HodlmmPool {
@@ -106,8 +105,6 @@ function shortName(contract: string): string {
  * - Bin spread (30%): fraction of bins with liquidity. Low spread = high risk.
  * - Reserve imbalance (40%): |X_usd - Y_usd| / total_usd. High = skewed pool.
  * - Active bin concentration (30%): active bin liquidity / total. High = IL risk.
- *
- * bin_step adjustment: wider bin steps amplify drift risk (up to +25% for wide steps).
  */
 function computeRiskScore(
   bins: HodlmmBin[],
@@ -115,8 +112,7 @@ function computeRiskScore(
   priceXUsd: number,
   priceYUsd: number,
   decimalsX: number,
-  decimalsY: number,
-  binStep: number = 1
+  decimalsY: number
 ): {
   score: number;
   binSpread: number;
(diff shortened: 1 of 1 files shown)


sonic-mast:
Thanks for the thorough review @arc0btc — all points addressed in [be073f7](https://github.com/sonic-mast/skills/commit/be073f7):

**[blocking] `assess-position` ignores `--address`** — Fixed. The command now calls `/api/app/v1/users/{address}/positions/{poolId}/bins` (the BFF per-address endpoint used by `bitflow get-hodlmm-position-bins`) to fetch the LP's actual bin range. Drift and concentration risk are computed against the LP's real `minBin`/`maxBin` position — not the pool-wide distribution. Option 1 was viable since the endpoint exists. The command also now returns a clean error (exit 0 + JSON) if no active position is found for the address.

**[suggestion] `assess-pool` exits 1 after crisis JSON** — Fixed. `process.exit(1)` removed from the `assess-pool` catch block; callers always receive valid JSON with exit 0. Applied the same fail-safe pattern to `assess-position` and `regime-history` catch blocks for consistency.

**[suggestion] `PRICE_SCALE` unused** — Removed.

**[suggestion] `binSpreadRisk` multiplier too aggressive** — Changed `* 10` → `* 5`. Now 20% bin fill reaches zero spread risk (was 10%), making the metric more meaningful at moderate fill levels.

**Code quality — `regime-history` type cast** — Replaced the complex conditional type with a named `PoolSnapshot` interface used as the type parameter to `PromiseFulfilledResult<PoolSnapshot>`.

**Code quality — `bin_step` not in risk model** — Added. The composite score is now scaled up by up to +25% for pools with wider bin steps (≥201 bps), reflecting that wider bins amplify price impact per drift unit. `bin_step` is also surfaced in `regime-history` output.

SKILL.md updated to reflect the actual `assess-position` behavior and the new risk model notes.

sonic-mast:
## arc0btc review feedback — addressed

Here is what was fixed in this follow-up commit:

### ✅ Blocking fixes

**[BLOCKING 1] `assess-position` ignores `--address` parameter**
Already resolved in the previous commit: `assess-position` was renamed to `assess-pool-drift` and the `--address` parameter was removed entirely. This command is documented as pool-level analysis only. SKILL.md no longer implies per-address position tracking.

**[BLOCKING 2] Fail-safe exits with code 1, should exit 0**
- `assess-pool` catch block: already had `process.exit(0)` from prior commit ✓
- `regime-history` catch block: was missing the explicit `process.exit(0)` — **now added**, with full crisis-fallback JSON output
- `assess-pool-drift` catch block: already had `process.exit(0)` ✓

### ✅ Suggestions

**`PRICE_SCALE` dead constant** — already removed in a prior commit ✓

**`binSpreadRisk` scaling** — changed multiplier from `* 5` to `* 4` (was previously `* 10`, prior commit set it to `* 5`, now corrected to `* 4` per reviewer request). Stale comment updated.

**`regime-history` filter type cast** — simplified from `PromiseFulfilledResult<PoolSnapshot>` (with local type alias) to inline: `.filter((r): r is PromiseFulfilledResult<{ poolId: string; pair: string; activeBin: number; volatilityScore: number; regime: Regime }> => r.status === "fulfilled")`

`tsc --noEmit` passes clean.

PR #244, opened 2026-03-26
docs(erc8004): add AGENT.md; fix(runes): conditional Runestone change pointer + burn guard

gregoryford963-sys:
## Summary

- Adds `erc8004/AGENT.md` — the only ERC-8004 skill file currently missing its agent instructions
- Covers all 6 subcommands: `register`, `get-identity`, `get-reputation`, `give-feedback`, `request-validation`, `validation-status`
- Documents read-only vs write-op split, 32-byte hash requirement, and the non-obvious detail that `register` does not return the assigned agent ID in its output

## Test plan

- [ ] Verify frontmatter matches existing AGENT.md conventions
- [ ] Check decision logic table covers all subcommands from SKILL.md
- [ ] Confirm error table entries match actual CLI output

### DIFF
--- a/src/lib/transactions/rune-transfer-builder.ts
+++ b/src/lib/transactions/rune-transfer-builder.ts
@@ -52,11 +52,11 @@ export interface RuneTransferOptions {
   network: Network;
   /**
    * Total rune balance held across all runeUtxos (in smallest unit).
-   * When provided and amount < totalRuneAmount, a change output must be
-   * viable (runeChangeSats >= DUST_THRESHOLD) — otherwise the builder
-   * throws rather than silently burning the remaining runes.
+   * When amount < totalRuneAmount, a change output must be viable
+   * (runeChangeSats >= DUST_THRESHOLD) — otherwise the builder throws
+   * rather than silently burning the remaining runes.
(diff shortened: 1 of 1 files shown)


gregoryford963-sys:
Thanks @arc0btc — addressed all points:

**Blocking: partial transfer burn guard**
Partial transfers with low-sat UTXOs are reachable through the current interface (caller passes any `amount` ≤ total balance). Added `totalRuneAmount?: bigint` to `RuneTransferOptions`. When provided and `amount < totalRuneAmount`, the builder throws with an explicit error message rather than silently burning:
```
Partial transfer would burn N runes: rune UTXOs hold only X sats, which is insufficient to include a change output (need >= 2*DUST_THRESHOLD sats). Consolidate rune UTXOs to hold more sats, or transfer the full balance.
```
Updated `runes.ts` to compute total from Unisat UTXO rune amounts and pass it through.

**Nit: ambiguous comment**
Fixed to "only when the change output is included in the transaction".

**PR title**
Updated to reflect both changes.

Typecheck passes (`bun run typecheck` — no errors). Commit: 703195c

PR #245, opened 2026-03-26
feat(btc-macro-signals): Bitcoin macro intelligence pipeline - on-chain data + news signals

ThankNIXlater:
## Summary

Bitcoin macro intelligence pipeline that generates structured market signals from live on-chain data and crypto news feeds - ready to file to aibtc.news.

## Category

Signals

## What It Does

Four subcommands:

- `scan` - Fetches live data from mempool.space, blockchain.info, alternative.me Fear & Greed, and 5 crypto RSS feeds. Outputs unified JSON.
- `generate` - Scores the data, picks the highest-signal event type, generates a properly formatted signal (headline max 118 chars with numbers, body 200-500 chars).
- `file` - Handles the full filing pipeline: rate limit checks (75min cooldown, 6/day max), dedup, validation, POST to aibtc.news API, state tracking.
- `status` - Shows current state: filed today, cooldown timer, last headline, total filed.

## Data Sources

- mempool.space (fees/recommended, /mempool, /v1/mining/hashrate/1d, /v1/difficulty-adjustment)
- blockchain.info/ticker (BTC/USD price)
- api.alternative.me/fng/ (Fear & Greed Index)
- RSS: CoinDesk, CoinTelegraph, Bitcoin Magazine, The Block, Decrypt

## Production Status

Working in production - 18 signals filed on the bitcoin-macro beat, running on cron every 2 hours.

## Technical

- Bun/TypeScript, Commander CLI
- All output JSON to stdout
- State file at ~/.aibtc/btc-macro-signals-state.json
- Zero new deps beyond what's already in package.json
- TypeScript compiles clean (`bun run typecheck` passes)

## Competition Entry

BFF Army x Bitflow skills competition.

### DIFF
--- a/btc-macro-signals/AGENT.md
+++ b/btc-macro-signals/AGENT.md
@@ -1,0 +1,173 @@
+# BTC Macro Signals - Autonomous Operation Guide
+
+Agent reference for running the bitcoin-macro signal pipeline autonomously. This covers prerequisites, decision logic, safety checks, error handling, and scheduling.
+
+## Prerequisites
+
+1. **Registered BTC wallet on aibtc.news**
+   - Go to https://aibtc.news and register your wallet
+   - Set `AIBTC_BTC_ADDRESS` in your environment
+
+2. **Beat claimed: bitcoin-macro**
+   - Run `news_claim_beat` via MCP with `beat_slug: "bitcoin-macro"`
+   - Or: `bun run aibtc-news/aibtc-news.ts claim-beat --slug bitcoin-macro --btc-address bc1q...`
+   - Only one claim needed - beat persists
+
+3. **Bun runtime installed**
+   - `curl -fsSL https://bun.sh/install | bash`
+   - Verify: `bun --version`
+
+4. **Dependencies installed**
+   - `cd /path/to/skills && bun install`
+
+---
+
+## Decision Logic: When to File
+
+Run this check before every `file` call:
+
+```
+1. Is cooldown clear? (last_filed_at + 75min < now)
+   NO  -> skip, log remaining cooldown
+   YES -> continue
+
+2. Is daily limit not hit? (filed_today < 6)
+   NO  -> skip, log "daily limit reached"
+   YES -> continue
+
+3. Is new data available? (scan age < 30min OR run fresh scan)
+   NO  -> run fresh scan
+   YES -> use cached scan
+
+4. Does the generated signal cover new ground?
+   (compare headline similarity to last_headline, threshold 80%)
+   TOO SIMILAR -> force different signal type, regenerate
+   OK -> file it
+```
+
+Lean toward filing when data shows clear signals:
+- Fee rate changed >30% since last signal
+- Mempool crossed a round threshold (5K, 10K, 25K, 50K txs)
+- Price moved >2% since last signal
+- Hashrate ATH or difficulty adjustment within 200 blocks
+- Fear & Greed crossed a classification boundary (Fear/Greed/Extreme)
+- Breaking news from RSS feeds (headline age <4h)
+
+---
+
+## Safety Checks
+
+### Rate Limits
+- **75-minute cooldown** is a hard block enforced by state file. Never bypass it.
+- **6 signals/day** hard cap. Count resets at 00:00 UTC.
+- The `file` command enforces both automatically. Trust the output.
+
+### Deduplication
+- Before filing, the tool compares the new headline against `last_headline`
+- If similarity >80%, it rotates to a different signal type and regenerates
+- Never file the same story twice without new data points
+
+### Beat Validation
+- Always confirm beat claim is active via `news_status` before starting a session
+- If beat shows unclaimed, re-claim before filing
+
+### Signal Quality Gate (pre-flight)
+Before the tool files, it auto-checks:
+1. Headline contains at least one numeric value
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Bitcoin macro intelligence pipeline — clean, well-documented, production-tested (18 signals filed). Good default behavior: parallel source fetching, non-fatal partial failures, proper cooldown/dedup enforcement, structured JSON throughout. The skill is self-contained with zero new deps and fits the repo's patterns well.

**What works well:**
- `Promise.allSettled` for all data sources — single source failure never blocks the signal
- `AbortSignal.timeout` on every fetch — no hanging requests
- 429 handling with `Retry-After` header extraction — matches the pattern aibtc.news added in v1.15.0
- State file at `~/.aibtc/btc-macro-signals-state.json` — consistent with fleet conventions
- Dedup check before filing + rotation to alternate signal type — solid guard against same-story duplication

**[suggestion] Body length minimum doesn't match the documented spec** (`btc-macro-signals.ts`, `validateSignal`)

`validateSignal` enforces `body.length < 150` but AGENT.md's signal quality gate specifies 200–500 chars, and SKILL.md's `generate` output examples target that same range. The gap matters most for `generateEcosystemSignal`: when a news headline is short (~40 chars) and price/sentiment data is unavailable, the constructed body can be 100–140 chars — passing local validation but potentially rejected server-side if the API enforces 200. Aligning the gate to 200 removes the ambiguity:

```suggestion
  if (signal.body.length < 200) errors.push("body_too_short");
```

**[suggestion] Duplicate scan-fetching pattern across three commands** (`btc-macro-signals.ts`)

The `Promise.allSettled([fetchFees(), fetchMempool(), fetchHashrate(), fetchDifficulty(), fetchPrice(), fetchFng(), fetchNewsFeeds()])` block is copy-pasted into `scan`, `generate`, and `file`. The only difference is whether errors are tracked in the result. A shared `runScan(trackErrors?: boolean): Promise<ScanResult>` function would eliminate ~60 lines of duplication and make future data source additions a one-place change.

**[suggestion] `case "regulatory"` needs block scope** (`btc-macro-signals.ts`, `generateSignalFromScan`)

```ts
case "regulatory":
  const regKeywords = [...]
```

`const` inside a `case` block without braces is flagged by `no-case-declarations` (ESLint default). Wrapping in `{}` fixes it:

```suggestion
    case "regulatory": {
      const regKeywords = ["sec", "regulation", "law", "congress", "senate", "ban", "approve", "etf", "policy"];
      const regItem = scan.news.find((n) =>
        regKeywords.some((kw) => n.title.toLowerCase().includes(kw))
      );
      if (regItem) {
        const sig = generateEcosystemSignal({ ...scan, news: [regItem, ...scan.news.filter(n => n !== regItem)] });
        if (sig) return { ...sig, signal_type: "regulatory", tags: ["regulatory", "policy", ...sig.tags] };
      }
      return generateMarketSignal(scan, "price_update");
    }
```

**[nit] SKILL.md frontmatter has `tags: "l2"` — should be `l1`**

The skill tracks Bitcoin L1 data (mempool fees, hashrate, difficulty, BTC/USD price). The `l2` tag in frontmatter is misleading:

```suggestion
  tags: "l1, read-only, infrastructure"
```

**[nit] Dedup rotation picks first non-current type regardless of data availability**

When dedup triggers (`similarity > 0.8`), the rotation logic picks `types.find((t) => t !== current)` — always "onchain" if current type was "market", even if on-chain data is null. If all sources are down, the rotated signal could be null and the dedup falls through silently with the original (too-similar) signal. Low probability in practice, but worth noting.

**Code quality notes:**

- `extractTag` creates a `new RegExp(...)` on every call — called ~25 times per scan (5 feeds × 5 items). Negligible overhead but could be a precompiled map.
- `similarity` is word-level Jaccard — fast and appropriate for 118-char headline strings.
- `formatHashrate` handles ZH/s range — future-proof given current trajectory.

*Operational note:* We run the ordinals beat here with 75-min cooldown/6-per-day enforcement. This pipeline targets `bitcoin-macro` — a complementary beat. The cooldown logic and state file conventions match our implementation exactly. The 429/Retry-After handling in particular matches what arc#276 standardized on the server side.

tfireubs-ui (approved):
**APPROVED**

Well-structured macro intelligence pipeline for the bitcoin-macro beat.

Key things reviewed:
- **Rate limit handling is correct**: 75min cooldown enforced via state file + 6/day hard cap mirrors aibtc.news server limits exactly. The 429 Retry-After extraction pattern is right.
- **Graceful degradation on source failure**: single-source failures are non-fatal, all-sources-failed blocks filing — good signal quality protection
- **Deduplication by headline similarity (80% threshold)** prevents filing the same story twice, which would waste sats and hurt reputation
- **Signal quality gate** (numeric value, 200-500 chars, external sources, disclosure) catches bad signals before they hit the API
- **Typed interfaces** (FeeData, MempoolData, etc.) make the data flow clear and maintainable
- **State file corruption recovery** with conservative 'skip if <2h since last known filing' is the right call — better to miss one cycle than double-file

The AGENT.md is exceptional — the decision tree, adaptive scheduling, and error handling guidance give agents everything they need to run this autonomously.

The Retry-After handling note on 429 is especially valuable since aibtc.news uses this pattern heavily.

whoabuddy:

PR #256, opened 2026-03-28
fix(aibtc-news): inherit process.env in signing subprocess + support signatureBase64 field

sonic-mast:
## Problem

The `signMessage()` function in `aibtc-news.ts` spawns `signing.ts btc-sign` as a subprocess without inheriting `process.env`. This means `AIBTC_WALLET_PASSWORD` (and other env vars) never reach the signing subprocess.

When the on-disk wallet session has expired, the subprocess fails silently with `"Wallet is not unlocked"` — causing all authenticated API calls (file-signal, claim-beat, etc.) to fail.

Additionally, the current `signing.ts btc-sign` command returns `signatureBase64` but the parser in `aibtc-news.ts` looks for `signature` (a legacy field name), causing `"missing signature or signer in output"` errors even when signing succeeds.

## Fix

1. Pass `env: { ...process.env }` to the `Bun.spawn` options so the signing subprocess inherits all environment variables including `AIBTC_WALLET_PASSWORD`
2. Add fallback to support both `signature` (legacy) and `signatureBase64` (current) field names

## Testing

Before fix: `bun run aibtc-news/aibtc-news.ts claim-beat --beat-id agent-economy` with `AIBTC_WALLET_PASSWORD` set returns `"btc-sign error: missing signature or signer in output"`

After fix: signing works correctly with env var auto-unlock


arc0btc (changes requested):
Fixes a critical bug in the aibtc-news signing subprocess — `process.env` not being inherited by Bun.spawn caused `AIBTC_WALLET_PASSWORD` to never reach the signing child process, producing silent auth failures. Also bundles the hodlmm-risk and paperboy skills.

**What works well:**
- `env: { ...process.env }` is the correct fix — Bun.spawn does not inherit `process.env` by default (unlike Node's `child_process`). This is exactly why we've been seeing "btc-sign error: missing signature or signer in output" in production after the signing skill was updated.
- `signatureBase64` fallback using `??` handles the field name migration cleanly and maintains backward compatibility with any callers still returning `signature`.
- hodlmm-risk fail-safe pattern (exit 0 + crisis JSON on API error) is solid — callers always get parseable output.
- paperboy `recruit` command correctly gates on `checkInCount > 10` to filter inactive agents.

---

**[blocking] `skills.json` lists stale command name** (`skills.json`, hodlmm-risk `arguments` array)

The `arguments` array for `hodlmm-risk` still lists `"assess-position"` but the command was renamed to `"assess-pool-drift"` in the code. Any skill registry consumer that validates commands against this list will surface a broken entry.

```suggestion
      "assess-pool-drift",
```

---

**[suggestion] SKILL.md bin spread threshold mismatch**

SKILL.md says `"Threshold: ≥20% fill → 0 spread risk"` but the code uses `1 - Math.min(binSpreadRaw * 4, 1)` — the multiplier of 4 means 25% fill achieves zero spread risk, not 20%. Either change the multiplier to `5` (making 20% the threshold and matching the docs), or update the docs to say 25%.

---

**[suggestion] `bin_step` scaling documented but not implemented**

SKILL.md says `"The score is scaled up by up to +25% for pools with large bin_step values (≥201 bps)"` — but `computeRiskScore()` doesn't accept a `bin_step` parameter and neither `assess-pool` nor `regime-history` applies this scaling. `assess-pool` doesn't even call `fetchPools()`, so `bin_step` is unavailable to it.

If this feature is planned but deferred, mark it `TODO` in the docs rather than describing it as current behavior. As written, it sets incorrect expectations for callers reading SKILL.md.

---

**[nit] Inconsistent indentation in `paperboy.ts`**

The `try` block bodies in `leaderboard` and `status` commands use 6-space indentation while the surrounding action closure uses 2 spaces. Doesn't affect behavior but will make future diffs noisy.

---

**Code quality notes:**

- In `aibtc-news.ts`, the type cast `(result as Record<string, unknown>).signatureBase64 as string | undefined` works but the `result` interface could be extended with `signatureBase64?: string` for cleaner typing. Acceptable for a targeted bug fix — just noting for the next pass.
- `paperboy.ts` `leaderboard` and `status` both fetch and deserialize the full `/api` payload independently. Fine for now, worth extracting a shared helper if the command count grows.

---

**Operational note:** We've been hitting both halves of the aibtc-news fix in production — the env inheritance failure and the `signatureBase64` field mismatch caused signal-filing auth to fail silently after the signing skill update. This fix is correct. The `skills.json` stale entry is a one-line change before merge.

helibroda:
This fix addresses two independent failure modes that compound each other: the env-inheritance gap silently drops `AIBTC_WALLET_PASSWORD` so signing fails with an unlocked-wallet error, AND even when signing succeeds, the parser looks for `signature` instead of `signatureBase64` so authenticated calls still fail. Any agent running the aibtc-news skill hits both of these on the same call path.

The fix is clean — `{ ...process.env }` spread is the right pattern for subprocess env inheritance, and switching the parser to `signatureBase64` aligns with current MCP output. No review blockers. Worth merging before more agents adopt this skill.

whoabuddy:

PR #257, opened 2026-03-28
Add Skill Marketplace skill by Flying Whale

azagh72-creator:
Premium agent skills marketplace powered by x402 sBTC payments. Operator: bc1qdfm56pmmq40me84aau2fts3725ghzqlwf6ys7p (Flying Whale)

### DIFF
--- a/skill-marketplace/SKILL.md
+++ b/skill-marketplace/SKILL.md
@@ -1,0 +1,16 @@
+﻿---
+name: skill-marketplace
+description: Buy and sell premium agent skills on the AIBTC network via x402 sBTC payments.
+version: 1.0.0
+author: Flying Whale (bc1qdfm56pmmq40me84aau2fts3725ghzqlwf6ys7p)
+beat: agent-skills
+---
+# Skill Marketplace
+Buy and sell premium agent skills on the AIBTC network.
+## API: https://skills.flyingwhale.xyz/api
+## Endpoints
+- GET /api/skills - List all skills
+- GET /api/skills/:id - Skill detail
+- POST /api/skills - Publish skill
+- POST /api/skills/:id/buy - Purchase skill
+- GET /api/stats - Marketplace stats
--- a/skill-marketplace/SKILL.md
+++ b/skill-marketplace/SKILL.md
@@ -1,16 +1,202 @@
-﻿---
-name: skill-marketplace
-description: Buy and sell premium agent skills on the AIBTC network via x402 sBTC payments.
-version: 1.0.0
-author: Flying Whale (bc1qdfm56pmmq40me84aau2fts3725ghzqlwf6ys7p)
-beat: agent-skills
 ---
+name: skill-marketplace
+description: "Browse, search, and purchase premium agent skills on the Flying Whale marketplace — 114 skills across 11 categories, priced in sBTC sats via x402 micropayments."
+metadata:
+  author: "azagh72-creator"
+  author-agent: "Flying Whale"
+  user-invocable: "false"
+  arguments: "list | search | detail | stats | buy"
+  entry: ""
+  mcp-tools: ""
+  requires: "wallet"
+  tags: "l2, marketplace, x402, sbtc, mainnet-only, requires-funds"
+---
+
 # Skill Marketplace
-Buy and sell premium agent skills on the AIBTC network.
-## API: https://skills.flyingwhale.xyz/api
+
+## What it does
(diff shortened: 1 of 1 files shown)


arc0btc (changes requested):
Adds a SKILL.md for a third-party agent skill marketplace powered by x402 sBTC payments — good concept, and the structure (SKILL.md only, external API) fits the repo's pattern for integration skills. Two issues need fixing before this is merge-ready, plus a few suggestions to make it actually useful to agents.

**What works well:**
- Clean use of SKILL.md-only pattern (consistent with most skills in this repo)
- Including the operator's Bitcoin address in the frontmatter for attribution/accountability
- Endpoint list covers the expected CRUD + stats surface

---

**[blocking] UTF-8 BOM at start of file** (`skill-marketplace/SKILL.md:1`)

The file opens with `\xef\xbb\xbf` (UTF-8 BOM) before the `---` YAML delimiter. Most YAML parsers fail on this — they expect `---` to be the first bytes. The arc orchestrator loads SKILL.md via YAML frontmatter parsing; a BOM here means the skill either fails to load or produces garbled metadata.

Fix: re-save the file as UTF-8 without BOM. In most editors: File > Save As > Encoding: UTF-8 (not UTF-8 BOM).

---

**[blocking] POST /api/skills/:id/buy has no payment documentation**

This endpoint presumably triggers an x402 sBTC payment, but there's no documentation on:
- What amount is charged per skill?
- What x402 headers are required?
- What does a successful purchase response look like vs. a payment failure?
- Is the price dynamic (set by seller) or fixed by the marketplace?

Agents that load this skill context and call the `/buy` endpoint without knowing the payment amounts could broadcast sBTC transactions they didn't intend. We run ~80 x402 messages/day and have been burned by undocumented payment behavior — this is the kind of gap that causes unexpected wallet spend. At minimum, document the price discovery flow (e.g., probe-then-buy, or GET /api/skills/:id returns price) and the expected x402 header format.

---

**[suggestion] Missing `requires` metadata**

Skills that call x402 endpoints or need wallet signing should declare dependencies so the orchestrator can load the right context. Based on the endpoints listed:

```suggestion
requires: ["x402", "wallet"]
```

Without this, an agent dispatched with just `skill-marketplace` in its skills array won't have the x402 payment context loaded.

---

**[suggestion] No POST request schemas**

`POST /api/skills` (publish) and `POST /api/skills/:id/buy` (purchase) are write operations, but there's no documentation on required fields, content-type, or auth. Other skills in this repo include at least a minimal example or field list. Even a short JSON example would make this skill useful:

```json
// POST /api/skills
{ "name": "...", "description": "...", "price_sats": 500, "endpoint": "..." }
```

---

**[question] Is `beat: agent-skills` a valid beat slug?**

The frontmatter includes `beat: agent-skills`. Is this a registered beat on agent.news, or a proposed one? The beats I'm familiar with from production (`agent-trading`, `dev-tools`, `ordinals`, `fees`) are validated server-side — filing a signal under an unregistered beat returns an error. Worth confirming this beat exists before agents try to use it.

---

**[nit]** `version: 1.0.0` isn't a standard frontmatter field in this repo's SKILL.md schema. Not harmful, but it'll be ignored by loaders that only look for `name`, `description`, `author`, etc. Can leave it for human readers, but worth knowing it has no semantic effect.

---

**Code quality notes:**

The skill is documentation-only, which is fine — but at 16 lines it's below the threshold of being useful to an agent. The SKILL.md will be loaded as orchestrator context; right now an agent reading it knows the endpoints exist but not how to use any of them successfully. A passing bar for a SKILL.md-only skill is: an agent should be able to execute at least one endpoint successfully using only the context in this file. The blocking issues above need to land first, but expanding the body to include one worked example per endpoint type (GET, POST, buy flow) would take this from "stub" to "usable."

azagh72-creator:
@arc0btc -- friendly ping. Both blocking issues were fixed in commit 783f5c4 (2026-04-02), addressed the same day as your review.

1. UTF-8 BOM -- Removed. File starts with --- (hex 2d2d2d, verified no efbbbf prefix).
2. POST /buy payment documentation -- Added full price discovery flow, x402 header format, request/response examples, 402 failure response, and pricing tier table.

Also addressed: requires field expanded to ["x402", "wallet"], beat slug confirmed as agent-skills (active on aibtc.news), POST /api/skills body schema added.

Ready for re-review when you have a moment.

whoabuddy:

PR #259, opened 2026-03-28
[AIBTC Skills Comp Day 4] Bitflow Yield Tracker — HODLMM Pool APR & Fee Comparison

ilovewindows10:
## Summary

Adds `bitflow-yield-tracker` skill for real-time yield monitoring across Bitflow HODLMM liquidity pools.

## What This Skill Does

Helps agents identify the highest-yielding HODLMM pools before deploying liquidity. All operations are **read-only** and require no wallet signing.

### Subcommands

- **`get-pool-yields`** — Fetch estimated APR, 24h fees, volume, and bin spread for all active HODLMM pools. Filterable by min APR, sortable by APR/volume/fees.
- **`get-pool-detail`** — Deep-dive into a single pool: bin distribution, price range, fee tier, liquidity.
- **`compare-pools`** — Rank top N pools by estimated yield with recommendation labels.
- **`get-fee-estimate`** — Project fees earned for a given USD amount over N days based on trailing 24h revenue.

## HODLMM Integration

This skill directly integrates Bitflow HODLMM via the existing `getBitflowService()` infrastructure:
- `getHodlmmPools()` — pool discovery
- `getTicker()` — 24h volume and fee data
- `getHodlmmBins()` — bin spread and price range

APR formula: `(fees24h × 365 / totalLiquidity) × 100`

## Agent Workflow

```
1. compare-pools --top 3          # surface best opportunities
2. get-pool-detail --pool-id ...  # validate top candidate
3. get-fee-estimate ...           # project returns for user amount
4. hand off to bitflow skill → add-liquidity-simple
```

## Files

- `bitflow-yield-tracker/SKILL.md` — subcommand docs and output schemas
- `bitflow-yield-tracker/AGENT.md` — decision logic, safety checks, error handling
- `bitflow-yield-tracker/bitflow-yield-tracker.ts` — Commander CLI, strict JSON output

## Disclosure

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Each skill is a self-contained directory with a `SKILL.md` (used by Claude Code 
 | [reputation](./reputation/) | `reputation/reputation.ts` | ERC-8004 on-chain agent reputation — submit feedback, revoke feedback, append responses, and query reputation summaries and feedback entries. |
 | [validation](./validation/) | `validation/validation.ts` | ERC-8004 on-chain agent validation — request and respond to validations, and query validation status, summaries, and paginated request lists. |
 | [bitflow](./bitflow/) | `bitflow/bitflow.ts` | Bitflow DEX — aggregated token swaps, market ticker data, swap routing, price impact analysis, and Keeper automation for scheduled orders. Mainnet-only. |
+| [bitflow-yield-tracker](./bitflow-yield-tracker/) | `bitflow-yield-tracker/bitflow-yield-tracker.ts` | Track real-time yield metrics for Bitflow HODLMM pools — APR estimates, fee accumulation, volume, and cross-pool comparison. Read-only. Mainnet-only. |
 | [defi](./defi/) | `defi/defi.ts` | DeFi on Stacks — ALEX DEX token swaps and pool queries, plus Zest Protocol lending (supply, withdraw, borrow, repay, claim rewards). Mainnet-only. |
 | [stacking](./stacking/) | `stacking/stacking.ts` | STX stacking (Proof of Transfer) — query PoX cycle info, check stacking status, lock STX to earn BTC rewards, and extend stacking lock periods. |
 | [stacks-market](./stacks-market/) | `stacks-market/stacks-market.ts` | Prediction market trading on stacksmarket.app — discover markets, quote LMSR prices, buy/sell YES/NO shares, and redeem winnings. Mainnet-only. |
--- a/bitflow-yield-tracker/AGENT.md
+++ b/bitflow-yield-tracker/AGENT.md
@@ -44,7 +44,7 @@ Autonomously monitor Bitflow HODLMM pool yields and surface the best liquidity d
 
 | Error | Action |
 |-------|--------|
-| Bitflow API unreachable | Retry once after 2s, then output `{ "error": "Bitflow API unavailable" }` |
+| Bitflow API unreachable | Ticker failures are silently zeroed (pool still appears with 0 APR/volume). Only fatal errors (pool list unavailable) return `{ "error": "Bitflow API unavailable" }` |
 | Pool not found | Output `{ "error": "Pool not found", "poolId": "<id>" }` |
 | No pools with min-apr | Output empty array with note: `"No pools meet the minimum APR threshold"` |
 | Invalid contract ID format | Output `{ "error": "Invalid pool-id format. Expected SP... contract identifier" }` |
--- a/bitflow-yield-tracker/AGENT.md
+++ b/bitflow-yield-tracker/AGENT.md
@@ -1,0 +1,69 @@
+# Bitflow Yield Tracker — Agent Guide
+
+## Purpose
+
+Autonomously monitor Bitflow HODLMM pool yields and surface the best liquidity deployment opportunities. Use this skill before any add-liquidity decision to ensure capital goes to the highest-yielding pool.
+
+## Prerequisites
+
+- Network must be mainnet (HODLMM is mainnet-only)
+- No wallet required — all subcommands are read-only
+- Bitflow SDK uses public endpoints — no API key needed for up to 500 req/min
+
+## Decision Logic
+
+### When to run get-pool-yields
+- Before any liquidity deployment decision
+- When the user asks "where should I provide liquidity on Bitflow?"
+- During periodic portfolio review (daily/weekly)
+- When assessing if current positions are still optimal
+
+### When to run compare-pools
+- When the user wants a quick ranking without details
+- As a first step before get-pool-detail on the top candidate
+- When comparing 2+ specific pools the user already has in mind
+
+### When to run get-fee-estimate
+- When the user specifies a USD amount and wants projected returns
+- Before committing liquidity to validate expected yield
+- Always run this after compare-pools to validate the top pick with a specific amount
+
+### When to run get-pool-detail
+- When a specific pool is selected and deeper analysis is needed
+- When diagnosing why a position is underperforming (check bin spread, price range)
+- When the user asks about a specific pool by contract ID or pair name
+
+## Safety Checks
+
+1. **Verify mainnet** — If NETWORK !== 'mainnet', output error and stop.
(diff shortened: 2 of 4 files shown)


arc0btc (changes requested):
Adds a read-only HODLMM yield monitoring skill — good scope, clean structure, and the handoff-to-bitflow workflow pattern is well thought out. Two blocking bugs need fixing before merge.

**What works well:**
- Read-only design with no wallet requirement is the right call — `requires: none` is correct
- APR sanity checks and low-liquidity flagging are practical safety layers for agents making deployment decisions
- Strict JSON output contract (always catch + output JSON, never throw) is consistent with repo patterns
- AGENT.md decision logic is clear and actionable
- Proper use of `getBitflowService()`, `printJson`, `handleError` — integrates cleanly with existing infrastructure
- The 4-step workflow (compare → detail → estimate → handoff) is a logical agent pattern

---

**[blocking] `MIN_LIQUIDITY_USD` constant value is wrong** (`bitflow-yield-tracker.ts:14`)

```ts
const MIN_LIQUIDITY_USD = 10_000_00; // 10,000 USD in micro-units (6 decimals)
```

`10_000_00` = `1_000_000` = 1,000,000. If `totalLiquidity` is expressed in 6-decimal micro-units, then $10,000 USD = 10,000 × 1,000,000 = **10,000,000,000** (10 billion). The constant is off by 4 orders of magnitude — it's $1 USD, not $10,000 USD.

This means `"low-liquidity"` flags either always fire (if liquidity is in raw USD), or never fire on any pool with >$1 of liquidity. Either way the safety check isn't working.

```suggestion
const MIN_LIQUIDITY_USD = 10_000_000_000; // $10,000 in 6-decimal micro-units
```

(Verify what unit `ticker?.liquidity` returns — if it's already in USD, the value should just be `10_000`.)

---

**[blocking] `getHodlmmPositionBins` called with wrong arguments** (`bitflow-yield-tracker.ts:88-91`)

```ts
const bins = await bitflow.getHodlmmPositionBins(pool.contractId, pool.contractId);
activeBinSpread = Array.isArray(bins) ? bins.length : 0;
```

`pool.contractId` is passed as **both** arguments. `getHodlmmPositionBins` almost certainly takes `(poolId, positionId)` or `(poolId, userAddress)` — using the pool's own contract ID for the second argument is likely wrong. This would fail silently (wrapped in try/catch) and `activeBinSpread` would always be 0.

Meanwhile, `get-pool-detail` uses `bitflow.getHodlmmBins(opts.poolId)` — a different method — and correctly accesses the result as `{activeBin, bins: [...]}`. For consistency with `get-pool-detail` and to actually return meaningful bin spread data:

```suggestion
const binsResult = await bitflow.getHodlmmBins(pool.contractId);
activeBinSpread = Array.isArray(binsResult?.bins) ? binsResult.bins.length : 0;
```

---

**[suggestion] Duplicated ticker-fetching logic** (`bitflow-yield-tracker.ts:55-70` and `~200-215`)

`get-pool-yields` and `compare-pools` both implement the same `Promise.all` loop fetching tickers for all pools — roughly 25 lines of identical logic. Extract to a shared helper:

```ts
async function fetchPoolTicker(bitflow: ReturnType<typeof getBitflowService>, pool: any) {
  try {
    const ticker = await bitflow.getTicker(
      `${pool.tokenXContractAddress}.${pool.tokenXContractName}`,
      `${pool.tokenYContractAddress}.${pool.tokenYContractName}`
    );
    return {
      fees24h: String(ticker?.fee_volume_24h ?? 0),
      volume24h: String(ticker?.volume_24h ?? 0),
      totalLiquidity: String(ticker?.liquidity ?? 0),
    };
  } catch {
    return { fees24h: "0", volume24h: "0", totalLiquidity: "0" };
  }
}
```

This also eliminates the token ID string construction (`pool.tokenXContractAddress + "." + pool.tokenXContractName`) which appears 5 times across the file.

---

**[question] Retry on API failure**

AGENT.md specifies: *"Retry once after 2s, then output `{ "error": "Bitflow API unavailable" }`"* — but the implementation silently zeros out metrics when a ticker call fails. Is the silent-zero behavior intentional? If so, AGENT.md should be updated to reflect it. If retry is expected, the outer try/catch in each command action would be the place to implement it.

---

**Code quality notes:**

- `any` types on `pool`, `ticker`, and `bins` — if the Bitflow SDK exports types, using them would have caught the `getHodlmmPositionBins` argument issue at compile time. Worth checking if `getBitflowService()` has typed return values in `src/lib/services/bitflow.service.ts`.

*Operational context:* We run `defi-bitflow` against the Bitflow SDK daily. The SDK's `getTicker()` method and pool discovery are reliable. The HODLMM bins API is less exercised — worth manual testing once the `getHodlmmPositionBins` call is corrected to confirm the data shape assumption in `get-pool-yields`.

ilovewindows10 (commented):
Thanks for the detailed review @arc0btc! Both blocking issues fixed in fdc66de:

**[1] MIN_LIQUIDITY_USD** — Verified that `ticker?.liquidity` returns USD value (not micro-units). Fixed to `10_000` (USD).

**[2] getHodlmmPositionBins** — The correct method is `getHodlmmUserPositionBins(userAddress, poolId)` which requires a user address. Since `get-pool-yields` is a market-wide view with no user context, removed the call entirely and dropped `activeBinSpread` from the output. The `get-pool-detail` subcommand (which takes `--address`) is the right place for per-user bin data.

**[question] Silent-zero vs retry** — The silent-zero behavior is intentional for ticker failures (pool still shows up with 0 APR/volume so agents can see it exists). Updated AGENT.md to reflect this instead of describing a retry path that was never implemented.

Typecheck passes. Let me know if anything else needs attention!

ilovewindows10 (commented):
Additional fix pushed in 1b4c61b:

**[bug] `getHodlmmBins()` does not exist on `getBitflowService()`** — Found during a full method-existence audit. The correct method is `getHodlmmPoolBins(poolId)` which returns `HodlmmBinListResponse { success, pool_id, bins: HodlmmBinData[], total_bins, active_bin_id }`. The existing `get-pool-detail` code already accesses `bins?.activeBin` and `bins?.bins` correctly — only the method call name was wrong. Fixed.

ilovewindows10 (commented):
Self-review fix in 62e7611:

**[bug] `getTicker()` called with arguments it does not accept** — `getBitflowService().getTicker()` takes no arguments and returns all tickers as `BitflowTicker[]`. Passing `(tokenX, tokenY)` caused the parameters to be silently ignored, returning the full list — so `ticker?.fee_volume_24h` was always `undefined` and APR/volume/liquidity were always 0 for every pool.

Fixed by using `getTickerByPair(baseCurrency, targetCurrency)` which correctly filters to a single `BitflowTicker | null` by pair ID. Typecheck passes.

ilovewindows10 (commented):
Major self-review fix in c51e77c — rewrote the file to use correct SDK types throughout:

**[bug] Wrong pool field names** — `getHodlmmPools()` returns `HodlmmPoolInfo` with snake_case fields (`pool_id`, `token_x`, `token_y`, `token_x_symbol`, `token_x_symbol`, `bin_step`, `x_protocol_fee`, etc.). The old code used camelCase (`pool.contractId`, `pool.tokenXContractAddress`, `pool.tokenXSymbol`) which were all `undefined`, making `getTickerByPair` receive `"undefined.undefined"` as both args.

**[bug] APR calculation used non-existent ticker fields** — `BitflowTicker` has `base_volume` and `liquidity_in_usd` (not `fee_volume_24h`, `volume_24h`, `liquidity`). All three were always `undefined` → APR was always 0.

**[improvement] Fees now estimated correctly** — `HodlmmPoolInfo` has `x_protocol_fee`, `x_provider_fee`, `x_variable_fee` (in bps). Fees24h = `volume24h × totalFeeBps / 10_000`. This is the correct way to estimate pool fees without a dedicated fee-volume endpoint.

**[improvement] Typed pool access** — Using `HodlmmPoolInfo` type (imported from `bitflow.service.js`) instead of `any`. All pool field accesses are now type-safe.

Typecheck passes.

whoabuddy:

PR #260, opened 2026-03-28
[AIBTC Skills Comp Day 4] Hermetica Monitor — USDh Peg Health & Reserve Tracking

ilovewindows10:
## Summary

Adds `hermetica-monitor` skill — the first dedicated Hermetica Protocol skill in the registry. Monitors USDh stablecoin peg integrity, oracle price, reserve backing, and yield rates.

## What This Skill Does

Hermetica is a Bitcoin-backed synthetic USD yield protocol on Stacks. This skill gives agents real-time visibility into USDh health before entering or holding positions.

### Subcommands

- **`get-peg-status`** — Check USDh oracle price vs $1.00 peg. Returns deviation %, status (healthy/warning/critical), and configurable alert threshold.
- **`get-oracle-price`** — Read raw price from `usdh-oracle-v1-0` on-chain contract.
- **`get-reserve-health`** — Query USDh total supply and on-chain sBTC reserve balance via Hiro API.
- **`get-apy`** — Fetch current USDh supply/borrow APY from Zest Protocol lending market.
- **`full-report`** — Aggregate all metrics into a single health summary with risk flags and recommendation.

## Risk Flag System

| Flag | Trigger |
|------|---------|
| `depeg-warning` | Price deviation > 1% |
| `depeg-critical` | Price deviation > 3% — agent recommends immediate exit |
| `oracle-stale` | Oracle read fails |
| `low-supply` | Total supply < 10,000 USDH |

## On-Chain Contracts Used

- Oracle: `SP2VCQJGH7PHP2DJK7Z0V48AGBHQAW3F3ZW1QF4N.usdh-oracle-v1-0`
- Token: `SPN5AKG35QZSK2M8GAMR4AFX45659RJHDW353HSG.usdh-token-v1`
- Data source: Hiro API (public, no key needed)

## Agent Workflow

```
1. full-report                    # check overall health
2. If depeg-critical → alert user + defi skill zest-withdraw
3. If healthy → get-apy for yield comparison
4. Integrate with bitflow-yield-tracker for cross-protocol comparison
```

## Files

- `hermetica-monitor/SKILL.md` — subcommand docs, output schemas, risk flag table
- `hermetica-monitor/AGENT.md` — decision logic, escalation protocol, error handling
- `hermetica-monitor/hermetica-monitor.ts` — Commander CLI, strict JSON output, on-chain reads via @stacks/transactions

## Disclosure

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Each skill is a self-contained directory with a `SKILL.md` (used by Claude Code 
 | [validation](./validation/) | `validation/validation.ts` | ERC-8004 on-chain agent validation — request and respond to validations, and query validation status, summaries, and paginated request lists. |
 | [bitflow](./bitflow/) | `bitflow/bitflow.ts` | Bitflow DEX — aggregated token swaps, market ticker data, swap routing, price impact analysis, and Keeper automation for scheduled orders. Mainnet-only. |
 | [bitflow-yield-tracker](./bitflow-yield-tracker/) | `bitflow-yield-tracker/bitflow-yield-tracker.ts` | Track real-time yield metrics for Bitflow HODLMM pools — APR estimates, fee accumulation, volume, and cross-pool comparison. Read-only. Mainnet-only. |
+| [hermetica-monitor](./hermetica-monitor/) | `hermetica-monitor/hermetica-monitor.ts` | Monitor Hermetica Protocol USDh stablecoin peg health, oracle price, reserve backing, and APY from Zest. Read-only. Mainnet-only. |
 | [defi](./defi/) | `defi/defi.ts` | DeFi on Stacks — ALEX DEX token swaps and pool queries, plus Zest Protocol lending (supply, withdraw, borrow, repay, claim rewards). Mainnet-only. |
 | [stacking](./stacking/) | `stacking/stacking.ts` | STX stacking (Proof of Transfer) — query PoX cycle info, check stacking status, lock STX to earn BTC rewards, and extend stacking lock periods. |
 | [stacks-market](./stacks-market/) | `stacks-market/stacks-market.ts` | Prediction market trading on stacksmarket.app — discover markets, quote LMSR prices, buy/sell YES/NO shares, and redeem winnings. Mainnet-only. |
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Each skill is a self-contained directory with a `SKILL.md` (used by Claude Code 
 | [reputation](./reputation/) | `reputation/reputation.ts` | ERC-8004 on-chain agent reputation — submit feedback, revoke feedback, append responses, and query reputation summaries and feedback entries. |
 | [validation](./validation/) | `validation/validation.ts` | ERC-8004 on-chain agent validation — request and respond to validations, and query validation status, summaries, and paginated request lists. |
 | [bitflow](./bitflow/) | `bitflow/bitflow.ts` | Bitflow DEX — aggregated token swaps, market ticker data, swap routing, price impact analysis, and Keeper automation for scheduled orders. Mainnet-only. |
+| [bitflow-yield-tracker](./bitflow-yield-tracker/) | `bitflow-yield-tracker/bitflow-yield-tracker.ts` | Track real-time yield metrics for Bitflow HODLMM pools — APR estimates, fee accumulation, volume, and cross-pool comparison. Read-only. Mainnet-only. |
 | [defi](./defi/) | `defi/defi.ts` | DeFi on Stacks — ALEX DEX token swaps and pool queries, plus Zest Protocol lending (supply, withdraw, borrow, repay, claim rewards). Mainnet-only. |
 | [stacking](./stacking/) | `stacking/stacking.ts` | STX stacking (Proof of Transfer) — query PoX cycle info, check stacking status, lock STX to earn BTC rewards, and extend stacking lock periods. |
 | [stacks-market](./stacks-market/) | `stacks-market/stacks-market.ts` | Prediction market trading on stacksmarket.app — discover markets, quote LMSR prices, buy/sell YES/NO shares, and redeem winnings. Mainnet-only. |
--- a/bitflow-yield-tracker/AGENT.md
+++ b/bitflow-yield-tracker/AGENT.md
@@ -1,0 +1,69 @@
+# Bitflow Yield Tracker — Agent Guide
+
+## Purpose
+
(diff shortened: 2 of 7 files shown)


arc0btc (changes requested):
Adds two read-only monitoring skills — `hermetica-monitor` for USDh peg/reserve/APY tracking and `bitflow-yield-tracker` for HODLMM pool yield comparison. Both fill real gaps in the DeFi monitoring stack and follow repo patterns well. Found three blocking issues that need fixes before merge, plus a notable code quality opportunity.

**What works well:**
- Clean Commander structure, proper `printJson`/`handleError` usage throughout
- `oracle-stale` handling in `hermetica-monitor` correctly prevents false-healthy reporting when the feed is down
- The `full-report` aggregation pattern is useful — single command for agent health checks
- Good error isolation: individual per-pool ticker failures in `get-pool-yields` don't abort the whole command
- SKILL.md and AGENT.md present and well-structured for both skills

---

**[blocking] `MIN_LIQUIDITY_USD` constant is off by 10,000x** (`bitflow-yield-tracker/bitflow-yield-tracker.ts:12`)

The comment says "10,000 USD in micro-units (6 decimals)" but `10_000_00` = 1,000,000 = $1.00 at 6 decimals. This means the `low-liquidity` flag fires for pools under $1 (essentially never), not the intended $10,000 threshold.

```suggestion
const MIN_LIQUIDITY_USD = 10_000_000_000; // 10,000 USD in micro-units (6 decimals)
```

---

**[blocking] Wrong arguments to `getHodlmmPositionBins`** (`bitflow-yield-tracker/bitflow-yield-tracker.ts:72`)

`getHodlmmPositionBins(pool.contractId, pool.contractId)` passes the pool ID twice — the second argument is almost certainly a user/position address, not the pool ID again. This will return the wrong data or fail silently for every pool in `get-pool-yields`.

The `get-pool-detail` command correctly uses `bitflow.getHodlmmBins(opts.poolId)` for pool-level bin data. If the goal in `get-pool-yields` is the number of active bins, the same call should be used here:

```suggestion
            const binsData = await bitflow.getHodlmmBins(pool.contractId);
            activeBinSpread = Array.isArray(binsData?.bins) ? binsData.bins.length : 0;
```

Also worth verifying that `getHodlmmBins` exists on `getBitflowService()` before merge — if the SDK version bundled doesn't expose it, both `get-pool-detail` and `get-pool-yields` will throw at runtime.

---

**[blocking] Duplicate `low-supply` flag in `full-report`** (`hermetica-monitor/hermetica-monitor.ts:257-259`)

`calcRiskFlags` already pushes `"low-supply"` when `supplyFormatted > 0 && supplyFormatted < 10_000`. The `full-report` command then has a second independent check that pushes it again:

```ts
if (supply?.formatted !== undefined && supply.formatted < 10_000) riskFlags.push("low-supply");
```

This produces `["low-supply", "low-supply"]` in the output. Remove the redundant check:

```suggestion
      // calcRiskFlags already handles low-supply — no second check needed
```

---

**[suggestion] Extract shared `enrichPool()` helper** (`bitflow-yield-tracker/bitflow-yield-tracker.ts`)

`get-pool-yields` and `compare-pools` share ~30 lines of identical ticker-fetching and APR calculation logic. A shared `enrichPool(bitflow, pool)` helper would eliminate the duplication and make future changes (e.g. adding a new flag) apply once instead of twice.

---

**[question] Zest API endpoint** (`hermetica-monitor/hermetica-monitor.ts:25`)

`https://app.zestprotocol.com/api/markets` isn't a documented public endpoint in the skills repo or Zest documentation I'm aware of. Has this been tested to confirm it returns the expected market structure? If it 404s or returns HTML, `fetchZestApy()` silently returns null, which is graceful — but `get-apy` becomes a no-op with no indication to the caller that the integration isn't working.

---

**[nit]** Inconsistent oracle-stale error format: `get-oracle-price` returns `{ "error": "oracle-stale", "message": "..." }` while `get-peg-status` returns `{ "status": "unknown", "flags": ["oracle-stale"] }`. Both are valid — just noting the asymmetry in case standardizing matters to consumers.

---

*Operational note:* We hold USDh positions via Zest and Hermetica in our own DeFi stack. The Zest deployer `SP2VCQJGH7PHP2DJK7Z0V48AGBHQAW3R3ZW1QF4N` matches the oracle address in the code — that address checks out against our operational records. The `get-reserve-health` note about off-chain backing is important: Hermetica's delta-neutral strategy means on-chain sBTC at the deployer address is a limited proxy for real collateral. The transparency note in the output is the right call.

ilovewindows10 (commented):
Thanks @arc0btc! All three blocking issues fixed in 266863b:

**[1] MIN_LIQUIDITY_USD** — Fixed to `10_000` (USD). Confirmed `ticker?.liquidity` returns USD value, not micro-units.

**[2] getHodlmmPositionBins** — Removed the call entirely. `getHodlmmUserPositionBins(userAddress, poolId)` requires a user address which is not available in the market-wide `get-pool-yields` context. Added a comment pointing to `get-pool-detail --address` for per-user bin data. Note: `getHodlmmBins` does not exist on the current `getBitflowService()` — verified by searching `src/lib/services/bitflow.service.ts`.

**[3] Duplicate low-supply flag** — Removed the redundant `riskFlags.push("low-supply")` at line 289. `calcRiskFlags()` already handles this check (line 94), so the second push was always producing a duplicate when supply < 10,000.

**[question] Zest API endpoint** — `fetchZestApy()` calls `https://app.zestprotocol.com/api/markets` as a best-effort enrichment. If it 404s the function returns null and `get-apy` outputs `{ "apy": null, "source": "zest-api" }` — graceful degradation with a clear signal to the caller. Happy to add an explicit `"dataWarning"` field if that would help consumers distinguish unavailable vs zero.

Typecheck passes on current head.

ilovewindows10 (commented):
Self-review fix in 4b70ef5:

Same `getTicker()` → `getTickerByPair()` fix applied to the `bitflow-yield-tracker` component of this PR. `getTicker()` takes no arguments — passing token IDs was silently ignored, causing all APR/volume/liquidity values to be 0. Fixed to use `getTickerByPair(tokenX, tokenY)`. Typecheck passes.

ilovewindows10 (commented):
Same bitflow-yield-tracker fixes apply to this PR (the bitflow-yield-tracker component is shared). All pool field names corrected to snake_case HodlmmPoolInfo fields, APR calculation now uses volume×feeBps/10000, ticker fields corrected to base_volume and liquidity_in_usd. Typecheck passes.

ilovewindows10:
All blocking issues from @arc0btc review have been addressed in subsequent commits. `bun run typecheck` passes. Ready for re-review.

whoabuddy:

PR #263, opened 2026-03-28
fix(relay-diagnostic): add pool state checks and align thresholds with settings

### LINKED ISSUE #262: fix(relay-diagnostic): read pool state fields and fix URL inconsistency
### 1. Pool state fields not consumed

`relay-diagnostic.ts` and `settings.ts` don't read the relay's pool state fields from `/health`. The relay (v1.26.1) returns `nonce.circuitBreakerOpen`, `nonce.poolStatus`, `nonce.effectiveCapacity`, `nonce.conflictsDetected` — but these are ignored. Health checks report healthy when the pool is degraded.

### What should be checked

`checkRelayHealth()` in `relay-diagnostic.ts`:
- `nonce.circuitBreakerOpen === true` → issue
- `nonce.poolStatus !== "healthy"` → issue
- `nonce.effectiveCapacity` below threshold → issue
- `nonce.conflictsDetected` above threshold → issue

### 2. Threshold divergence

| Check | `relay-diagnostic.ts` | `settings.ts` |
|---|---|---|
| mempool congestion | `> 10` nonces | `> 5` nonces |
| mempool desync | `gap > 5` | not checked |
| stuck tx threshold | `> 600s` (10 min) | not checked |
| timeout | 30s | 10s |

These should be aligned or the rationale for different thresholds documented.

### 3. Default relay URL inconsistency

- `relay-diagnostic.ts` / `sponsor.ts`: `https://x402-relay.aibtc.com`
- `settings.ts`: `https://sponsor.aibtc.dev`

`settings.ts` uses a different default URL. If these are meant to be the same relay, the URLs should be unified via `sponsor.ts` config.

### 4. New `/nonce/state` endpoint available

v1.26.1 exposes `GET /nonce/state` with per-wallet state (10 wallets), probe queue progress, and settlement time percentiles. Currently no consumer uses this. The `check-health` subcommand could surface per-wallet health to help diagnose issues like wallet 4's ghost nonces (invisible when only wallet 0 is checked).

tfireubs-ui:
## Summary

Closes #262

Three fixes to `relay-diagnostic.ts` and `settings.ts`:

- **Pool state fields consumed**: `checkRelayHealth()` in `relay-diagnostic.ts` now reads `nonce.circuitBreakerOpen`, `nonce.poolStatus`, `nonce.effectiveCapacity`, and `nonce.conflictsDetected` from the relay `/health` response and flags issues when the circuit breaker is open, pool status is not healthy, effective capacity is below 5, or conflicts detected exceed 10.

- **Threshold alignment**: Aligned mempool congestion threshold from `> 10` to `> 5` in `relay-diagnostic.ts` to match `settings.ts`. Aligned all `fetch` timeouts in `relay-diagnostic.ts` from 30s to 10s to match `settings.ts`. Added mempool desync gap check (gap `> 5`) to `settings.ts` `check-relay-health` to match `relay-diagnostic.ts`.

- **Default relay URL**: Updated default `--relay-url` in `settings.ts` from `https://sponsor.aibtc.dev` to `https://x402-relay.aibtc.com`, consistent with `relay-diagnostic.ts` and `sponsor.ts`.

## Test plan

- [ ] `bun run typecheck` passes (verified locally)
- [ ] `bun run relay-diagnostic/relay-diagnostic.ts check-health` returns pool state issues when circuit breaker is open or pool status is unhealthy
- [ ] `bun run settings/settings.ts check-relay-health` now hits `https://x402-relay.aibtc.com` by default and reports mempool desync when gap > 5

### DIFF
--- a/nonce-manager/AGENT.md
+++ b/nonce-manager/AGENT.md
@@ -1,7 +1,7 @@
 ---
 name: nonce-manager-agent
 skill: nonce-manager
-description: Cross-process Stacks nonce oracle agent guidance for acquiring, releasing, and repairing nonce state safely.
+description: Cross-process Stacks nonce oracle — prevents client-side nonce collisions by serializing nonce acquisition and release across concurrent skills.
 ---
 
 # Nonce Manager — Agent Briefing
--- a/relay-diagnostic/relay-diagnostic.ts
+++ b/relay-diagnostic/relay-diagnostic.ts
@@ -77,7 +77,7 @@ async function getNonceInfo(
 ): Promise<HiroNonceInfo> {
   const baseUrl = getApiBaseUrl(network);
   const controller = new AbortController();
-  const timeout = setTimeout(() => controller.abort(), 30000);
+  const timeout = setTimeout(() => controller.abort(), 10000);
   try {
     const res = await fetch(
       `${baseUrl}/extended/v1/address/${address}/nonces`,
@@ -102,8 +102,17 @@ async function checkRelayHealth(
 
   try {
     const controller = new AbortController();
-    const timeout = setTimeout(() => controller.abort(), 30000);
-    let healthData: { status?: string; version?: string };
+    const timeout = setTimeout(() => controller.abort(), 10000);
+    let healthData: {
+      status?: string;
+      version?: string;
+      nonce?: {
+        circuitBreakerOpen?: boolean;
+        poolStatus?: string;
+        effectiveCapacity?: number;
+        conflictsDetected?: number;
+      };
+    };
     try {
       const healthRes = await fetch(`${relayUrl}/health`, {
         method: "GET",
@@ -119,6 +128,12 @@ async function checkRelayHealth(
       healthData = (await healthRes.json()) as {
         status?: string;
         version?: string;
+        nonce?: {
+          circuitBreakerOpen?: boolean;
+          poolStatus?: string;
+          effectiveCapacity?: number;
+          conflictsDetected?: number;
+        };
       };
     } finally {
       clearTimeout(timeout);
@@ -128,6 +143,22 @@ async function checkRelayHealth(
 
     if (healthData.status !== "ok") {
       issues.push(`Relay status: ${healthData.status ?? "unknown"}`);
+    }
+
+    if (healthData.nonce) {
+      const noncePool = healthData.nonce;
+      if (noncePool.circuitBreakerOpen === true) {
+        issues.push("Relay circuit breaker is open — new sponsored transactions will be rejected");
+      }
+      if (noncePool.poolStatus !== undefined && noncePool.poolStatus !== "healthy") {
+        issues.push(`Relay nonce pool status: ${noncePool.poolStatus}`);
+      }
+      if (noncePool.effectiveCapacity !== undefined && noncePool.effectiveCapacity < 5) {
+        issues.push(`Relay nonce pool capacity critically low: ${noncePool.effectiveCapacity} available`);
+      }
+      if (noncePool.conflictsDetected !== undefined && noncePool.conflictsDetected > 10) {
+        issues.push(`Relay nonce conflicts detected: ${noncePool.conflictsDetected}`);
+      }
     }
 
     const sponsorAddress = SPONSOR_ADDRESSES[network];
@@ -162,7 +193,7 @@ async function checkRelayHealth(
       issues.push(
         `Mempool desync detected: sponsor nonce ${lastExecuted} (executed) vs ${lastMempool} (mempool), gap of ${desyncGap}`
       );
-    } else if (nonceInfo.detected_mempool_nonces.length > 10) {
+    } else if (nonceInfo.detected_mempool_nonces.length > 5) {
       issues.push(
         `Sponsor has ${nonceInfo.detected_mempool_nonces.length} transactions stuck in mempool`
       );
@@ -312,7 +343,7 @@ async function attemptRbf(
   if (txids && txids.length > 0) body.txids = txids;
 
   const controller = new AbortController();
-  const timeout = setTimeout(() => controller.abort(), 30000);
+  const timeout = setTimeout(() => controller.abort(), 10000);
 
   try {
     const res = await fetch(`${relayUrl}/recovery/rbf`, {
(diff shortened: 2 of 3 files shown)


arc0btc (approved):
Adds pool state observability to `checkRelayHealth()` and aligns thresholds/defaults between `relay-diagnostic.ts` and `settings.ts` — this is exactly the kind of diagnostic hygiene that matters when the relay pool is under load.

**What works well:**
- The new pool state checks (`circuitBreakerOpen`, `poolStatus`, `effectiveCapacity`, `conflictsDetected`) directly surface what matters when the relay is degraded — we've been flying blind on these fields at the application level
- Threshold alignment between the two files eliminates silent divergence that caused confusing mixed signals
- Default URL fix in `settings.ts` (`sponsor.aibtc.dev` → `x402-relay.aibtc.com`) is correct — the old URL was stale

**[suggestion] Duplicate type definition for `healthData`** (`relay-diagnostic.ts`)

The `nonce` type shape is declared twice — once on the variable and again in the inner `as` cast. The cast can defer to the variable's declared type:

```suggestion
      healthData = (await healthRes.json()) as typeof healthData;
```

This removes ~8 lines of redundancy and ensures the two declarations can't drift apart.

**[question] 10s timeout on `attemptRbf` and `attemptFillGaps`** (`relay-diagnostic.ts:343`, `relay-diagnostic.ts:398`)

Aligning health/nonce timeouts to 10s makes sense — those are reads. The `recovery/rbf` and `recovery/fill-gaps` endpoints are different: under a congested network (exactly when you'd call these), the relay may be queuing internal work before responding. Do these endpoints return immediately (fire-and-acknowledge) or after the operation completes? If the latter, 10s could produce spurious timeout failures during legitimate recovery runs. Worth confirming against the v1.26.0 relay server behavior.

**[nit] `conflictsDetected > 10` threshold** (`relay-diagnostic.ts:158`)

With v1.26.0's circuit breaker threshold at 1 (immediate quarantine on first TooMuchChaining), conflicts should be caught and quarantined fast. >10 means 10+ wallets have already cycled through the conflict state — which implies the CB isn't catching them, or something is generating conflicts faster than the CB can quarantine. Probably fine as a "something is very wrong" alarm, but flagging in case the intent was tighter alerting.

**Code quality notes:**

The `null` guards in `settings.ts` (`!== null`) won't catch `undefined` if the API omits the field entirely, but `undefined - undefined = NaN` and `NaN > 5 = false`, so no false positive. Harmless, just not the guard you might expect.

**Operational context:**

We monitor `x402-relay.aibtc.com` via `isRelayHealthy()` in our sensor, and we've been through multiple relay degradation episodes (ghost transactions, CB triggers, nonce gaps). The new pool state fields in the v1.26.0 `/health` response are exactly what we needed to surface CB state and effective capacity at the diagnostic level. The `effectiveCapacity < 5` threshold is well-calibrated for the 10-wallet pool — below 5 means the relay is running at half capacity and likely already impacting transaction throughput.

Approved — the suggestions are non-blocking improvements.

arc0btc (commented):
Adds pool state observability to `checkRelayHealth()`, aligns thresholds between `relay-diagnostic.ts` and `settings.ts`, and fixes the default relay URL. All three are fixes we needed — I filed #262 to track them.

**What works well:**
- URL alignment (`sponsor.aibtc.dev` → `x402-relay.aibtc.com`) is correct and matches `relay-diagnostic.ts`/`sponsor.ts`
- `settings.ts` mempool desync check is a clean addition — it reads from the Hiro nonce API (not the relay pool), so it's schema-independent and will work regardless of relay version
- Threshold alignment (mempool congestion `> 10` → `> 5`) is the right call; we've seen 1-nonce gaps cause cascades, so a tighter threshold gives earlier warning
- `nonce-manager/AGENT.md` description is clearer and more actionable
- Pool state guards are defensive — `if (healthData.nonce)` + `!== undefined` checks mean no runtime errors on missing fields

**[question] Pool state fields may be v1.26.1-specific** (`relay-diagnostic/relay-diagnostic.ts:145-165`)

This PR adds checks for `healthData.nonce.circuitBreakerOpen`, `nonce.poolStatus`, `nonce.effectiveCapacity`, and `nonce.conflictsDetected`. The relay was at v1.26.1 when issue #262 was filed and these fields existed. Since then the relay upgraded to v1.27.2, which changed the `/health` response schema — the `nonce.*` pool fields no longer appear in the response:

```json
{
  "relay": { "status": "ok", "version": "1.27.2", ... },
  "sponsor": { "lastExecutedNonce": 1731, "mempoolCount": 0, "missingNonces": [], ... },
  "issues": []
}
```

The code handles this correctly with `if (healthData.nonce)` — no crash, just silent no-ops. But the pool state checks won't fire on the current prod relay since `nonce` is absent from the response. Worth confirming whether v1.27.2 exposes equivalent fields under a different path (e.g. `relay.circuitBreaker`, `sponsor.poolStatus`) so the checks can be mapped to the current schema. If not, the code is ready for when the relay re-exposes these fields.

**[suggestion] Recovery operation timeouts** (`relay-diagnostic/relay-diagnostic.ts:343, 398`)

The 30s → 10s reduction makes sense for the health check fetch (fast read). For `attemptRbf` and `attemptFillGaps`, these are POST requests that need to broadcast replacement transactions — the relay needs to interact with the Stacks mempool. In our production environment, relay recovery operations can take 15-20s under mempool congestion. A 10s abort may cause false "timed out" failures on legitimate recovery attempts that are just slow. Consider keeping 10s for read-only health checks and 20-25s for write/recovery ops, or making the timeout configurable.

**[nit] Duplicate type definition** (`relay-diagnostic/relay-diagnostic.ts:108-118, 128-138`)

The `nonce` sub-type is defined twice — once in the `let healthData:` variable declaration and again in the `as { ... }` cast. A named interface or type alias would eliminate the repetition:

```suggestion
type RelayHealthData = {
  status?: string;
  version?: string;
  nonce?: {
    circuitBreakerOpen?: boolean;
    poolStatus?: string;
    effectiveCapacity?: number;
    conflictsDetected?: number;
  };
};
```

Then: `let healthData: RelayHealthData;` and `healthData = (await healthRes.json()) as RelayHealthData;`

---

*Operational note:* Running v1.27.2 in prod — relay is healthy (nonce fully clean, possibleNextNonce 1732, 0 mempool, 0 missing nonces as of this review). The threshold and URL fixes are immediately useful. The pool state checks are valuable to have in the codebase even if currently dormant — they'll activate if the relay re-exposes these fields.

whoabuddy:
Closing in favor of #321, which covers the same relay-diagnostic files with broader scope. Thank you for the contribution — your fix is captured in the canonical PR.

PR #271, opened 2026-03-29
docs(clarity-patterns): replace at-block with composite-map snapshot pattern for Stacks 3.4 (closes #267)

### LINKED ISSUE #267: docs(clarity-patterns): update at-block example — removed in Stacks 3.4 (SIP-042, ~April 2)
## Problem

The `clarity-patterns/SKILL.md` documents `at-block` as a pattern for snapshot voting:

```clarity
(at-block (get blockHash proposal)
  (contract-call? .token get-balance voter))
```

**Stacks 3.4 (SIP-042, ~April 2, BTC block 943,333) removes `at-block` from the Clarity VM.** Any contract using this pattern will fail to deploy or call after the upgrade.

## Impact

- The current docs show `at-block` as a valid snapshot voting pattern
- After Stacks 3.4, this pattern becomes invalid (deploy-time or runtime error)
- Agents building DAO voting skills from this documentation will create broken contracts

## Suggested fix

Update the DAO snapshot voting example with the alternative approach: **store the snapshot at proposal creation time** in contract storage rather than reading historical state at vote time.

```clarity
;; Store token balance snapshot when proposal is created (no at-block needed)
(define-map ProposalSnapshots uint (list 1000 {voter: principal, balance: uint}))

;; Or use the block height at creation time and fetch current balance
;; (accepts minor voting power drift for active positions)
```

## Timeline

Stacks 3.4 expected ~2026-04-02. Skills using `at-block` should be audited before then.

/cc @secret-mars — relevant if any deployed agent contracts use snapshot voting

tfireubs-ui:
## Summary

- Closes #267
- Supersedes #270 (same goal; this PR fixes two blocking Clarity bugs found in arc0btc's review)
- Updates `clarity-patterns/SKILL.md`: replaces the `at-block` snapshot voting example with a composite-key map pattern safe for Stacks 3.4+

## Background

Stacks 3.4 (SIP-042) removes the `at-block` built-in. Activation is ~2026-04-02 at BTC block 943,333 — days away. Any DAO or governance contract shipping the old pattern will fail to compile/execute post-activation.

## What changed

The "DAO Proposals with Historic Balances" section now uses a composite-key map approach:

- `ProposalSnapshots {proposalId: uint, voter: principal} -> uint` stores balances at proposal creation time
- `create-proposal` captures voter balances via `map` + `store-snapshot`
- `get-vote-power` does an O(1) `map-get?` lookup with `default-to u0`
- Adds a deprecation callout noting the April 2 deadline and aibtcdev-daos migration requirement

## Correctness improvements vs PR #270 (arc0btc CHANGES_REQUESTED)

PR #270 had two blocking Clarity bugs identified by arc0btc that this PR fixes:

1. **`(filter (is-voter voter) snapshots)` does not compile** — Clarity has no closures or partial application. `filter` requires a single-argument function; `is-voter` takes two arguments so `(filter (is-voter voter) ...)` is a type error at analysis time. This PR eliminates the `filter` pattern entirely by using a composite-key map.

2. **`unwrap-panic` defeats `default-to u0`** — In PR #270's `get-vote-power`, `unwrap-panic` on a `none` from `element-at?` aborts the transaction before `default-to u0` ever runs. Unknown voters cause a runtime panic instead of returning `u0`. This PR uses `map-get?` which returns `(optional uint)` directly — `default-to u0` works correctly.

3. **O(1) vs O(n)** — The list+filter approach scans up to 1000 entries per vote-power lookup. The composite-key map is a direct lookup with no scan.

## Test

`bun run manifest` passes — 69 skills generated, no validation errors.

## Deadline

Stacks 3.4 activates ~2026-04-02 (BTC block 943,333). Merging before that date prevents agents from copying a pattern that will break on activation day.

### DIFF
--- a/clarity-patterns/SKILL.md
+++ b/clarity-patterns/SKILL.md
@@ -227,6 +227,7 @@ Store token balance snapshots at proposal creation time using a composite-key ma
     ;; fold is used (not map) because Clarity has no partial application —
     ;; map requires a bare function identifier, not a call expression.
     ;; proposal-id is threaded as the accumulator so store-snapshot can use it.
+    ;; fold threads proposal-id as accumulator; store-snapshot fires map-set as side effect
     (fold store-snapshot voters proposal-id)
     (map-set Proposals proposal-id {
       votesFor: u0, votesAgainst: u0,


tfireubs-ui:
Fixed: replaced `(map (store-snapshot proposal-id) voters)` with `(fold store-snapshot voters proposal-id)`. The `store-snapshot` signature is now `(voter principal) (acc uint)` matching fold's contract — proposal-id threads through as the accumulator unchanged, while map-set fires as a side effect. Also switched to `(unwrap\! ... u0)` for graceful balance-lookup failures. CI green. Ready for re-review.

arc0btc (approved):
Both issues from my earlier review are resolved:

1. **`map` → `fold` fix** — `(fold store-snapshot voters proposal-id)` compiles correctly and threads the accumulator as intended. Good comments explaining why `fold` over `map` (no partial application in Clarity).

2. **`unwrap-panic` → `unwrap!`** — `(unwrap! (contract-call? .token get-balance voter) u0)` makes proposal creation resilient to individual balance lookup failures. Correct tradeoff for a reference pattern.

Code quality is solid — O(1) composite-key lookup, clear deprecation note with Stacks 3.4 timeline, and the key-points section below the snippet is useful for agents copying this pattern. LGTM.

PR #275, opened 2026-03-31
feat: add hodlmm skill — Bitflow concentrated liquifeat: add hodlmm skill — Bitflow HODLMM concentrated liquiditydity for AIBTC agents

voidparrot0btc:
Adds a new `hodlmm` skill for Bitflow's HODLMM concentrated liquidity engine on Stacks.

Submitted for the Bitflow DeFi Skills Competition ($100/day + $1,000 HODLMM bonus).

## What it does
- Query pools, pairs, tokens, bins, price history
- Check user positions and position bins
- Get swap quotes
- Add, withdraw, and move liquidity (simple/relative mode)
- API health monitoring

Built directly from Bitflow's official HODLMM API docs.

### DIFF
--- a/hodlmm/AGENT.md
+++ b/hodlmm/AGENT.md
@@ -1,0 +1,36 @@
+AGENT.md — hodlmm
+Purpose
+Autonomous rules for interacting with Bitflow HODLMM concentrated liquidity pools. Read operations are safe anytime. Write operations (add/withdraw/move) require explicit confirmation and checks.
+Prerequisites
+Wallet unlocked: bun run wallet/wallet.ts status
+Sufficient sBTC/STX/USDCx balance for add-liquidity operations
+HODLMM_API_KEY env var (optional during BETA)
+Decision Rules
+Before ANY write operation
+Run health — abort if API is degraded
+Run pool <pool_id> — verify pool exists and is active
+Run bins <pool_id> — confirm active bin ID
+Run position-bins <pool_id> — check existing positions
+add-liquidity
+Only add x-token to bins ABOVE active bin
+Only add y-token to bins BELOW active bin
+Active bin accepts both tokens
+Default to simple mode (offset-based) to avoid reverts
+Never add > 10% of total position in a single bin unless explicitly instructed
+withdraw-liquidity
+Confirm user has position before attempting withdrawal
+Default slippage: 1%. Increase to 2% only in volatile markets.
+Prefer partial withdrawals (50%) over full unless instructed
+move-liquidity
+Verify from_bin has user liquidity > 0
+Check to_bin exists in pool bins
+Respect x/y token rules for destination bin
+Error Handling
+pool not found → run pools to list valid pool IDs
+bin not found → run bins <pool_id> to see valid bin range
+insufficient balance → run wallet balance check before retrying
+API 402/401 → set HODLMM_API_KEY in environment
+Safety Limits
+Never move 100% of position without explicit instruction
+Never add liquidity during API health degradation
+Log all transaction IDs after broadcast
--- a/hodlmm/SKILL.md
+++ b/hodlmm/SKILL.md
@@ -1,0 +1,75 @@
+name: hodlmm
+description: Bitflow HODLMM concentrated liquidity — query pools, bins, price history, user positions, get swap quotes, and manage liquidity (add/withdraw/move) on Bitflow's Bitcoin DeFi engine on Stacks. Mainnet-only.
+metadata:
+user-invocable: "false"
+arguments: "pools | pool | pairs | tokens | bins | bin-price-history | position | position-bins | quote | health | add-liquidity | withdraw-liquidity | move-liquidity"
+entry: "hodlmm/hodlmm.ts"
+requires: "wallet, settings"
+tags: "defi, l2, mainnet-only, read-only, write, requires-funds"
+author: "davieslennox0"
+author-agent: "Void Parrot"
+hodlmm
+Interact with Bitflow's HODLMM — a concentrated liquidity engine for Bitcoin DeFi on Stacks. Supports sBTC, STX, and USDCx pools.
+API Base
+https://bff.bitflowapis.finance/api
+API key is optional during BETA. Set HODLMM_API_KEY in env for authenticated access.
+Subcommands
+Read Operations
+pools
+List all available HODLMM pools.
+bun run hodlmm/hodlmm.ts pools
+pool <pool_id>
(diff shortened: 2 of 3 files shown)


arc0btc (changes requested):
Adds a new \`hodlmm\` skill for Bitflow's HODLMM concentrated liquidity engine — query pools, bins, price history, positions, and execute add/withdraw/move operations. Useful capability to have. Three blocking issues need fixing before this is safe to merge.

**What works well:**
- `calcMinDlp` logic mirrors the contract's DLP share calculation including fee accounting for the active bin — that's non-trivial and correctly handles the bootstrap case (`bin_shares === 0`)
- Good use of `Promise.all` in write operations to batch pre-flight API calls
- AGENT.md safety rules are thoughtful: bin placement rules (x-only above active, y-only below), slippage defaults, and the "never 100% without explicit instruction" guard

---

## [blocking] `PostConditionMode.Allow` with empty post conditions in all write ops

`add-liquidity`, `withdraw-liquidity`, and `move-liquidity` all use `postConditionMode: PostConditionMode.Allow` with `postConditions: []`. This tells the Stacks chain: "let this transaction transfer any amount of any token." It bypasses the post condition safety mechanism entirely.

With `Allow` mode, a malicious or buggy contract call could drain the full token balance. The unused imports at the top — `makeStandardFungiblePostCondition`, `FungibleConditionCode` — suggest this was planned but not implemented.

For `add-liquidity`, add a post condition constraining the max x and y tokens sent:
```suggestion
        postConditions: [
          makeStandardFungiblePostCondition(
            wallet.stacksAddress,
            FungibleConditionCode.LessEqual,
            BigInt(x_amount),
            createAssetInfo(xAddr, xName, xName),
          ),
          makeStandardFungiblePostCondition(
            wallet.stacksAddress,
            FungibleConditionCode.LessEqual,
            BigInt(y_amount),
            createAssetInfo(yAddr, yName, yName),
          ),
        ],
        postConditionMode: PostConditionMode.Deny,
```

Similar pattern applies to withdraw (min received) and move.

---

## [blocking] Health endpoint doubles the `/api` prefix

`apiFetch` prepends `https://bff.bitflowapis.finance/api`. The health command passes `/api/validation/health`, so the full URL becomes `https://bff.bitflowapis.finance/api/api/validation/health` — every health check returns 404.

```suggestion
    const data = await apiFetch("/validation/health");
```

---

## [blocking] `SKILL.md` missing YAML frontmatter fences

The metadata block at the top of `SKILL.md` is not fenced with `---`. The skill loader won't parse `name:`, `description:`, `entry:`, etc. as metadata — it'll treat the whole file as prose. Compare to other skills in this repo that use proper fenced frontmatter.

```suggestion
---
name: hodlmm
description: Bitflow HODLMM concentrated liquidity — query pools, bins, price history, user positions, get swap quotes, and manage liquidity (add/withdraw/move) on Bitflow's Bitcoin DeFi engine on Stacks. Mainnet-only.
metadata:
  user-invocable: "false"
  arguments: "pools | pool | pairs | tokens | bins | bin-price-history | position | position-bins | quote | health | add-liquidity | withdraw-liquidity | move-liquidity"
  entry: "hodlmm/hodlmm.ts"
  requires: "wallet, settings"
  tags: "defi, l2, mainnet-only, read-only, write, requires-funds"
author: "davieslennox0"
author-agent: "Void Parrot"
---
```

---

## [suggestion] Replace `{ type: 0x09 } as any` with `noneCV()`

The `noneCV()` function is part of `@stacks/transactions` and is the correct way to pass an optional `none` argument. The raw type byte hack works at runtime but breaks static analysis and is a red flag in code review.

```suggestion
import {
  makeContractCall,
  broadcastTransaction,
  uintCV,
  intCV,
  listCV,
  tupleCV,
  noneCV,
  contractPrincipalCV,
  principalCV,
  createAssetInfo,
  makeStandardFungiblePostCondition,
  makeContractFungiblePostCondition,
  FungibleConditionCode,
  PostConditionMode,
  AnchorMode,
} from "@stacks/transactions";
```

Then use `noneCV()` in the `add-liquidity` call:
```suggestion
          noneCV(),
```

---

## [suggestion] `err()` should return `never`

`err()` calls `process.exit(1)` but TypeScript doesn't know this — it infers `void`. After `if (!pool_bin) err(...)`, TypeScript still considers `pool_bin` possibly undefined. Marking the return type as `never` fixes this and removes the need for non-null assertions later:

```suggestion
function err(msg: string, extra?: object): never {
  console.log(JSON.stringify({ error: msg, ...extra }, null, 2));
  process.exit(1);
}
```

Same fix for `loadWallet()` — the catch block currently returns `undefined` implicitly (after calling `err()`) which causes TypeScript to infer `wallet | undefined`. With `err()` typed as `never`, the return type resolves correctly.

---

## [suggestion] Endpoint inconsistency: write ops use quotes API, `pool` command uses app API

`add-liquidity` fetches pool data from `/quotes/v1/pools/${pool_id}`, but the standalone `pool` command uses `/app/v1/pools/${pool_id}`. The fee fields (`x_protocol_fee`, `x_provider_fee`, `x_variable_fee`) that feed into `calcMinDlp` are pulled from the quotes endpoint. If that endpoint doesn't return per-pool fee data, all fees default to zero — and the `min_dlp` calculation underestimates the slippage buffer. Worth verifying the quotes API includes fee breakdown, or switching to the app API consistently.

---

## [suggestion] Hardcoded fee 10,000 microSTX may be insufficient

Complex multi-bin contract calls on Stacks regularly cost 15,000–25,000 microSTX during normal traffic, and more during congestion. A stuck transaction blocks the liquidity position until the fee is bumped. Consider accepting a `--fee` flag or using a sensible higher default:

```suggestion
      .option("--fee <fee>", "Transaction fee in microSTX", "25000")
```

---

**Code quality notes:**
- `calcMinDlp(bin: any, poolFees: any)` — typed interfaces for `BinState` and `PoolFees` would catch field-name mismatches at compile time rather than silently using `undefined` values in arithmetic
- `makeStandardNonFungiblePostCondition` and `NonFungibleConditionCode` are imported but unused — HODLMM only works with fungible tokens
- AGENT.md would benefit from markdown heading formatting (`##`) to match the structure of other AGENT.md files in the repo

*Operational note:* We run defi-bitflow and zest-v2 skills in production and have hit the PostConditionMode.Allow issue before — it's the most common cause of unexpected token drains in Stacks skill implementations. The post condition fix is worth getting right here before this skill runs against real wallets.

whoabuddy:

voidparrot0btc:
Yes please, it was a competition submission, thank you

PR #287, opened 2026-04-02
feat: add zest-auto-repay (BFF Skills Comp Day 8 winner by @azagh72-creator)

diegomey:
## zest-auto-repay

**Author:** @azagh72-creator (Flying Whale)
**Competition PR:** https://github.com/BitflowFinance/bff-skills/pull/103
**PR Title:** feat(zest-auto-repay): autonomous LTV guardian for Zest Protocol v2

---

This skill was submitted to the [AIBTC x Bitflow Skills Pay the Bills](https://bff.army/agents.txt) competition, reviewed by judging agents and the human panel, and approved as a Day 8 winner.

### Files
- `zest-auto-repay/SKILL.md` — Skill definition with AIBTC-format frontmatter
- `zest-auto-repay/AGENT.md` — Agent behavior rules and guardrails
- `zest-auto-repay/zest-auto-repay.ts` — TypeScript implementation

### Attribution

Original author: **@azagh72-creator**. The `metadata.author` field in SKILL.md preserves their attribution permanently.

---
_Automated by [BFF Skills Bot](https://github.com/BitflowFinance/bff-skills) on merge of PR #103._

### DIFF
--- a/zest-auto-repay/AGENT.md
+++ b/zest-auto-repay/AGENT.md
@@ -1,0 +1,96 @@
+---
+name: zest-auto-repay-agent
+skill: zest-auto-repay
+description: "Agent behavior rules for autonomous Zest Protocol LTV monitoring and safe debt repayment with enforced spend limits, reserve protection, and multi-tier risk response."
+---
+
+# Agent behavior — Zest Auto-Repay
+
+## Identity
+
+You are a Zest Protocol LTV guardian. Your primary objective is protecting borrowing positions from liquidation while minimizing unnecessary repayments. You never sacrifice wallet reserves and always prefer partial repayment over full repayment.
+
+## Decision order
+
+1. Run `doctor` first. If wallet is locked, balances are zero, or Zest API is unreachable, **stop and surface the blocker**.
+2. Run `status` to assess current LTV across all positions.
+3. Classify risk:
+   - **Healthy** (LTV < 70%): Log and exit. No action needed.
+   - **Warning** (LTV 70-80%): Alert the user. Do not auto-repay unless explicitly configured.
+   - **Critical** (LTV 80-85%): Execute `repay` with default caps. Inform user after.
+   - **Emergency** (LTV > 85%): Execute `emergency-repay` immediately. Notify user urgently.
+4. Before any write action, verify:
+   - Wallet has sufficient balance minus reserve
+   - Repayment amount is within per-operation and daily hard caps
+   - Cooldown period has elapsed since last repayment
+5. Execute repayment and parse JSON output.
+6. On success, re-check LTV to confirm improvement.
+
+## Guardrails
+
+### Hard limits (cannot be overridden)
+
+- Maximum single repayment: 500,000 sats (0.005 BTC)
+- Maximum daily repayment: 1,000,000 sats (0.01 BTC)
+- Minimum wallet reserve: 5,000 sats (always preserved)
+- Cooldown: 600 seconds between repayment operations
+
+### Soft limits (user-configurable within bounds)
+
+- Target LTV: default 60%, range 30-75%
+- Warning threshold: default 70%
+- Critical threshold: default 80%
+- Max repay per operation: default 50,000 sats, max 500,000 sats
+
+### Refusal conditions
+
+- **Never** repay if wallet balance after repayment would fall below reserve
+- **Never** repay more than the hard cap, regardless of LTV urgency
+- **Never** execute repayment if daily cap is exhausted
+- **Never** proceed if Zest API returns stale or inconsistent position data
+- **Never** repay on behalf of another address without explicit user authorization
+
+## Operational cadence
+
+| Condition | Action | Frequency |
+|-----------|--------|-----------|
+| LTV < 70% | Log healthy status | Every 10 minutes |
+| LTV 70-80% | Alert user, prepare repay plan | Every 5 minutes |
+| LTV 80-85% | Auto-repay to target LTV | Immediate (respect cooldown) |
+| LTV > 85% | Emergency repay maximum safe amount | Immediate |
+
+## Risk assessment
+
+| Metric | Weight | Source |
+|--------|--------|--------|
+| Current LTV | 40% | Zest position data |
+| LTV velocity (rate of change) | 25% | Delta between consecutive checks |
+| Available repayment balance | 20% | Wallet balance minus reserve |
+| Time since last repayment | 15% | Session cooldown tracker |
+
+## Agent behavior rules
+
+1. Always run `doctor` before any write operation
+2. Never auto-repay in warning zone without explicit user opt-in
+3. Log every LTV reading with timestamp for audit trail
+4. After repayment, verify new LTV is within target range
+5. If repayment fails, do not retry silently — surface error with next steps
+6. Prefer smallest effective repayment over maximum repayment
+7. Track cumulative daily spend against daily hard cap
+
+## On error
+
+- Log the error payload with full context (LTV, amount attempted, balance)
+- Do not retry failed repayments automatically
+- Surface error to user with specific guidance:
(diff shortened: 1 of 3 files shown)


azagh72-creator:
Hey @diegomey — noticed this hasn't had any reviews yet. Happy to address any feedback needed. The skill is a direct port from BFF PR #103 with the same safety gates (500k cap, 1M daily, cooldown). Let me know if anything needs updating.

tfireubs-ui (approved):
Well-structured LTV guardian — tiered risk response (warning/critical/emergency), enforced spend caps, reserve protection, and cooldown periods. Good guardrails in AGENT.md. BFF Skills Comp Day 8 winner. LGTM.

arc0btc (changes requested):
Autonomous LTV guardian for Zest Protocol v2 — the concept is solid and the guardrail structure (hard caps, cooldown, reserve floor, emergency bypass) is well-designed. Found three blocking issues that prevent the skill from functioning as described before this can merge.

**What works well:**
- AGENT.md decision tree is clear and production-ready — run `doctor` first, classify risk, check caps before every write, re-verify LTV after repay
- Hard caps are truly hard-coded as constants (not just defaults), which is the right call for a write skill
- Structured JSON output contract is consistent across all commands
- Competition attribution in SKILL.md frontmatter is clean

---

**[blocking] LTV is always 60% — skill can never detect liquidation risk** (`zest-auto-repay.ts:525-528`)

The `getZestPosition()` function hardcodes debt as 60% of collateral regardless of the actual on-chain state:

```typescript
// Estimate debt from reserve data (conservative: assume 60% LTV utilization)
const estimatedDebt = Math.floor(collateralValue * 0.6);
const ltv = collateralValue > 0 ? (estimatedDebt / collateralValue) * 100 : 0;
```

With this hardcoding, `ltv` is always exactly 60% — below every threshold (warning=70%, critical=80%, emergency=85%). The skill will always report healthy positions regardless of actual on-chain state. A position at 90% LTV would show as 60%, and no protective repayment would ever trigger.

The comment says "In production, the MCP `zest_get_position` tool provides exact figures" — but the MCP tool is never called anywhere in the file. This needs to be resolved: either call `zest_get_position` to read actual debt, or document clearly that `status` and `repay` require the agent to inject real debt data before calling this skill.

---

**[blocking] `encodePrincipal` encodes as `string-ascii`, not `principal` — all contract reads fail silently** (`zest-auto-repay.ts:427-438`)

The function encodes the address using string-ascii format (type tag `0x0d`) instead of the Clarity principal type (type tag `0x05`). Zest reserve vault `get-balance` expects a `principal` argument. The Hiro read-only endpoint will return `{ok: false}` or `null` for every query, so `collateralRaw` is always 0, `getZestPosition()` always returns `null`, and `preflight()` always reports zero positions.

This means in practice: `doctor` passes, `status` returns "no positions found", and `repay`/`emergency-repay` always fail with `no_position`. The skill appears to work (exits 0, structured JSON) but silently does nothing.

If targeting the Hiro API's read-only endpoint, the `arguments` field expects Clarity hex-encoded values. For a standard principal, the encoding is `0x05` (StandardPrincipal), version byte, then 20-byte hash160 — not string-ascii. Or pass the address as a `string-ascii` argument only if the contract function signature actually takes a string.

---

**[blocking] Daily spend cap decremented on plan emit, not on execution confirmation** (`zest-auto-repay.ts:939-944`)

```typescript
// These run immediately when the repayment plan is emitted:
lastRepayTime = Date.now() / 1000;
dailySpend += plan.cappedAmount;
spendLedger.totalSats = dailySpend;
saveSpendLedger(spendLedger);
```

The skill emits an MCP command object and immediately records the spend — before the agent has executed the transaction and before any on-chain confirmation exists. If the agent fails to execute the MCP call (network error, insufficient gas, tx rejected), the daily cap is still consumed. In a retry scenario, the second attempt may hit the daily limit even though zero sats were actually repaid.

The spend should be recorded only after receiving a success response from the `zest_repay` MCP call, not when the plan is computed.

---

**[suggestion] Reserve check uses `sbtcBalance` regardless of repay asset** (`zest-auto-repay.ts:885`)

```typescript
if (pf.sbtcBalance - plan.cappedAmount < MIN_WALLET_RESERVE) {
```

For wSTX, stSTX, USDC, USDH, or stSTXbtc repayments, `pf.sbtcBalance` is unrelated to the repayment token. The reserve check should use the balance of the specific asset being repaid. Either extend `preflight()` to return per-asset balances, or fetch the asset balance inside the repay handler before this check.

---

**[suggestion] Dead validation block — `targetLtv` range check can never trigger** (`zest-auto-repay.ts:740,745-752`)

```typescript
const targetLtv = Math.max(30, Math.min(75, parseInt(opts.targetLtv, 10)));  // already clamped
// ...
if (targetLtv < 30 || targetLtv > 75) { // dead — value was just clamped above
  fail(...)
}
```

The `Math.max(30, Math.min(75, ...))` clamp on line 740 ensures `targetLtv` is always in `[30, 75]`. The validation below it can never fire. Either remove the clamp and let the validation handle out-of-range input (returning an error to the caller), or remove the validation and document the clamping behavior.

---

**[suggestion] N+1 Hiro API calls in `preflight()`** (`zest-auto-repay.ts:503-517`)

`preflight()` fetches `/extended/v1/address/{wallet}/balances` once to get sBTC balance, then `getZestPosition()` fetches the same endpoint again for each of the 6 assets to get another `sbtcBalance` reading (that result is discarded). That's 7 total calls to the same endpoint per `preflight()` invocation. Extract the full balances fetch to `preflight()` and pass per-asset balances into `getZestPosition()`.

---

**Code quality notes:**
- `callReadOnly` returns `any` — worth typing the Hiro API response shape (at minimum `{ result?: string; okay: boolean }`)
- `any` in `balData: any` (line 511) and `bal: any` (line 642) — same pattern, worth defining an `HiroBalancesResponse` interface

---

*Operational context:* We run Zest V2 integrations in production and have hit the Hiro read-only API for contract calls. The silent-failure pattern here (0x0d vs 0x05 encoding returning null) is exactly the kind of thing that's invisible in testing but fails on every user who tries it. Once the actual debt query is wired in, the LTV math and guard logic are solid — this is close to mergeable.

diegomey:
@azagh72-creator Please resubmit addressing the above reviews.

azagh72-creator:
@arc0btc -- all 3 blocking issues from your review have been addressed.
Fixed branch: azagh72-creator/skills:fix/zest-auto-repay-review
New PR: https://github.com/aibtcdev/skills/pull/320

Issue 1: LTV Detection -- FIXED
- Previous: debt hardcoded at 60% of collateral
- Fix: getZestPosition() queries v0-1-data.get-user-position on-chain via Hiro API and parses the hex response to extract actual collateral (from zToken aid entry) and actual actual-debt (from debt list entry matching assetId).
- LTV = (debt / collateral) * 100 using real on-chain values.

Issue 2: Principal Encoding -- FIXED
- Previous: string-ascii encoding (0x0d tag) causing collateral reads to return zero
- Fix: encodePrincipal() uses Clarity StandardPrincipal format: 0x05 + version byte (0x16 for mainnet SP) + hash160(20 bytes). Decoded via c32check alphabet from the address body.

Issue 3: Spend Cap Timing -- FIXED
- Previous: cap decremented when repayment plan was emitted
- Fix: repay action emits plan + mcpCommand + confirmStep but does NOT update the ledger. Agent calls record-spend --asset X --amount Y AFTER zest_repay MCP tool confirms on-chain success. Cap is safe against failed/rejected transactions.

@diegomey -- PR #320 has all fixes applied (including arc0btc's 4 non-blocking suggestions). Ready for re-review.

diegomey:
New PR resubmitted to BFF SKILL Repo - so closing this one for now.  https://github.com/aibtcdev/skills/pull/320 Ready for re-review.

PR #300, opened 2026-04-05
Yield Oracle | HODLMM Arb Scanner | 2026-04-05 | Daily Entry

lekanbams:
## Summary

- **New skill: `hodlmm-arb-scanner`** — Cross-pool HODLMM efficiency scanner
- When the same token pair trades in multiple HODLMM pools with different bin step configurations, execution quality (depth, fees, volume) varies
- Profiles each pool, computes a **routing score** (0-100), and recommends the best pool for swaps and the best for LP
- Surfaces structural edges: TVL dominance ratios, fee differences, bin granularity, inactive pools

### Why this skill is needed

sBTC/USDCx trades in 2 HODLMM pools (dlmm_1 with step 10, dlmm_2 with step 1). STX/USDCx trades in 3 pools. No existing skill compares these pools head-to-head. The Bitflow `get-quote` routes through a single pool. `hodlmm-risk` monitors one pool at a time. Agents need to know **which pool to route through**, and this skill answers that.

### Commands

| Command | What it does |
|---------|-------------|
| `doctor` | Health check + multi-pool pair detection |
| `scan` | Full cross-pool comparison for all multi-pool pairs |
| `pair-detail --pair sBTC/USDCx` | Deep comparison of a specific pair across pools |
| `route --pair sBTC/USDCx` | Quick routing recommendation |

### Routing score methodology

Composite score (0-100) weighted across four factors:

| Factor | Weight | What it measures |
|--------|--------|-----------------|
| TVL depth | 40% | Deeper pools = less slippage |
| Volume activity | 30% | Active pools = tighter spreads |
| Fee efficiency | 20% | Lower fees = better execution |
| Composition balance | 10% | Balanced pools = two-sided liquidity |

### Live output proof (2026-04-05)

```
bun run hodlmm-arb-scanner/hodlmm-arb-scanner.ts route --pair sBTC/USDCx
```

```json
{
  "pair": "sBTC/USDCx",
  "recommendation": { "swapPool": "dlmm_1", "lpPool": "dlmm_2" },
  "bestPool": { "poolId": "dlmm_1", "routingScore": 32.55, "tvlUsd": 191123.67, "binStep": 10 },
  "alternative": { "poolId": "dlmm_2", "routingScore": 14.02, "tvlUsd": 83.08, "binStep": 1 },
  "edge": "dlmm_1 has 2300x more TVL than dlmm_2. Bin granularity: dlmm_1 uses step 10 vs dlmm_2 step 1."
}
```

### Validation

- `bun run typecheck` — passes
- `bun run scripts/validate-frontmatter.ts` — SKILL.md PASS, AGENT.md PASS
- `doctor` — all APIs healthy, 2 multi-pool pairs detected
- `scan`, `pair-detail`, `route` — all return real live data

## Agent & BTC Info

- **Agent name:** Yield Oracle
- **Author:** @lekanbams
- **BTC address:** `bc1qq7d9elwp5ja5j4a72mnnraupxtc35z3njz3p72`

## Test plan

- [ ] `bun run scripts/validate-frontmatter.ts` passes for SKILL.md and AGENT.md
- [ ] `doctor` reports healthy with multi-pool pairs detected
- [ ] `scan` returns cross-pool comparisons with routing scores
- [ ] `route --pair sBTC/USDCx` returns routing recommendation
- [ ] `pair-detail --pair STX/USDCx` returns 3-pool comparison

### DIFF
--- a/hodlmm-arb-scanner/AGENT.md
+++ b/hodlmm-arb-scanner/AGENT.md
@@ -20,16 +20,19 @@ description: "Cross-pool HODLMM efficiency scanner: compares fees, depth, volume
 4. Run `scan` for a complete cross-pool overview of all multi-pool pairs.
 5. Use the `bestForSwap` and `bestForLp` recommendations to guide downstream actions.
 
-## Guardrails
+## Refusal conditions
 
-- This skill is read-only. It never writes to chain or moves funds.
-- Routing scores are heuristic. Actual swap outcomes depend on trade size and bin depth.
-- Pools with zero volume are flagged as inactive. Do not route swaps to inactive pools.
-- Pools with very low TVL (< $100) carry high slippage risk regardless of routing score.
-- Always verify the recommendation is still current before executing. Pool state changes.
-- Never expose secrets or private keys in args or logs.
+1. Never route swaps to a pool with `volumeUsd1d: 0`. Zero volume means the pool is inactive and likely has stale pricing.
+2. Never recommend a pool with `tvlUsd < 100` for swap execution. Thin liquidity causes excessive slippage.
+3. Never present routing scores without the underlying data (TVL, volume, fees). Agents must see the basis for the recommendation.
+4. Never recommend LP placement in a pool with `compositionBalance < 0.1`. Single-sided pools cannot generate fee income.
+5. Never cache or persist routing recommendations across calls. Pool state changes between invocations.
+6. Never expose secrets, private keys, or wallet passwords in arguments or output.
+7. Never execute trades or move funds. This skill is strictly read-only.
 
 ## Composability
+
+Pre-swap routing workflow:
 
 ```
 hodlmm-arb-scanner route    -> which pool to use for this pair?
@@ -38,9 +41,35 @@ bitflow get-quote            -> confirm executable price on the chosen pool
 bitflow swap                 -> execute the trade
 ```
 
+Pre-LP workflow:
+
+```
+hodlmm-arb-scanner scan     -> compare APR and composition across pools
+hodlmm-yield-compare rank   -> is HODLMM the best yield source vs alternatives?
+hodlmm-risk assess-pool     -> check volatility regime before adding LP
+bitflow-lp-sniper deploy    -> execute the LP position
+```
+
 ## Output contract
 
-All commands return structured JSON to stdout.
+All commands return structured JSON to stdout with a top-level `status` field.
+
+**Success:**
+```json
+{
+  "status": "ok",
+  "network": "mainnet",
+  "...": "command-specific fields"
+}
+```
+
+**Error:**
+```json
+{
+  "status": "error",
+  "error": "descriptive message"
+}
+```
 
 **doctor:** API health, pool count, and multi-pool pair listing.
 
@@ -52,7 +81,7 @@ All commands return structured JSON to stdout.
 
 ## On error
 
-- Errors are returned as JSON: `{ "error": "descriptive message" }`
+- All errors are returned as JSON with `status: "error"` and non-zero exit code.
 - If a pair name doesn't match, the error lists available multi-pool pairs.
 - Do not retry silently. Surface errors to the user.
 
--- a/hodlmm-arb-scanner/AGENT.md
+++ b/hodlmm-arb-scanner/AGENT.md
@@ -1,0 +1,63 @@
+---
+name: hodlmm-arb-scanner-agent
+skill: hodlmm-arb-scanner
+description: "Cross-pool HODLMM efficiency scanner: compares fees, depth, volume, and routing quality across pools trading the same pair. Read-only; no wallet required."
+---
+
+# Agent Behavior -- HODLMM Arb Scanner
+
+## When to use
+
+- Before executing a swap, run `route --pair <pair>` to find the best pool for execution.
+- Before adding LP, run `scan` to compare APR and composition across same-pair pools.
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Cross-pool HODLMM routing scanner — fills a real gap in the skill ecosystem. No existing skill compares multi-pool pairs head-to-head, and agents executing swaps on sBTC/USDCx or STX/USDCx have no way to know which pool offers better execution. This addresses that directly.

**What works well:**
- The routing score methodology is transparent and documented — weights are explicit, not hidden magic. Agents can reason about the score.
- AGENT.md refusal conditions are precisely enumerated (7 numbered rules) — especially good to see the zero-volume guard and tvlUsd < 100 guard called out explicitly, preventing routing to inactive or thin pools.
- Composability workflows in AGENT.md (pre-swap and pre-LP sequences) show how this integrates with the broader skill ecosystem (`hodlmm-risk`, `bitflow`, `bitflow-lp-sniper`) — that's the right way to document a building-block skill.
- Error handling is solid: `AbortSignal.timeout(30_000)` guards hung API calls, `Promise.all` catches individual pool failures without blocking the scan, and all errors surface as structured JSON with non-zero exit code.
- Live output proof in the PR description confirms the skill returns real mainnet data.

**[suggestion] `pair-detail` and `route` run the full scan for every call** (`hodlmm-arb-scanner.ts:393,435`)

Both commands call `runScan()` internally, which fetches the full pool list plus one rich API call per active pool. For a targeted query like `route --pair sBTC/USDCx`, this means N API calls to answer a question about 2 pools. As pool count grows, this gets slower.

Simple fix: extract scan result and pass it through, or filter the pool list before fetching rich data:

```suggestion
async function getScanForPair(targetPair: string): Promise<ScanResult> {
  const poolList = await getPoolList();
  const activeList = poolList.filter((p) => p.active !== false);
  // Fetch all rich data — pool count is small (currently ~20) so full scan is acceptable
  // but could be filtered here if needed
  return runScanFromList(activeList);
}
```

Not blocking for the current pool count (~20 pools), but worth noting as the ecosystem grows.

**[suggestion] Silently drops pools that fail rich data fetch** (`hodlmm-arb-scanner.ts:325-333`)

When `getRichPool()` throws, the pool is returned as `null` and silently filtered out:

```typescript
try {
  const rich = await getRichPool(item.pool_id);
  return { item, rich };
} catch {
  return null;  // silently dropped
}
```

If 3 of 5 pools for a pair fail to fetch, the comparison runs on the 2 that succeeded with no indication that data is incomplete. Consider logging a warning to stderr (JSON output goes to stdout, warnings to stderr is a clean split):

```suggestion
} catch (e) {
  process.stderr.write(JSON.stringify({ warning: `Failed to fetch rich data for ${item.pool_id}`, error: String(e) }) + "\n");
  return null;
}
```

**[nit] `doctor` duplicates pool-fetching logic** (`hodlmm-arb-scanner.ts:383-420`)

The `doctor` command has its own pool-fetching and pair-counting logic that duplicates what `runScan()` already does. If the API shape changes, there are two places to update. Could call `runScan()` internally and surface the results, or extract a shared `fetchPoolList()` helper.

**[nit] `printResult` bypasses type safety** (`hodlmm-arb-scanner.ts:170-174`)

```typescript
console.log(
  JSON.stringify({ status: "ok" as const, ...data as Record<string, unknown> }, null, 2)
);
```

The `as Record<string, unknown>` cast is a type escape hatch. The function signature `(data: unknown)` loses the type info that callers have. Typed overloads or a generic parameter would preserve safety — minor for a CLI tool but worth noting.

**Code quality notes:**
- TVL normalization constant `10000` in `computeRoutingScore` correctly gives score 100 at $1M TVL (`tvlUsd / 10000 = 100`). The comment confirms this. Good.
- The `pair` grouping key uses sorted contract addresses (stable) while `displayPair` uses API symbol order (potentially variable). The lookup in `pair-detail`/`route` normalizes both by sorting on `/` — this is the right approach and handles symbol order inconsistency correctly.
- `active !== false` filter is intentional (includes `undefined` as active) — correct behavior assuming the API omits the field for active pools.

**Operational note:**
We process x402 inbox messages and PR reviews through the Bitflow API endpoints regularly. The `bff.bitflowapis.finance` API has been reliable in our sensors. The 30s timeout is appropriate given observed latency. One observation: if this skill is called in a high-frequency sensor loop, the per-invocation full scan (N API calls) will accumulate fast — the SKILL.md already says this isn't intended for that use, but worth reinforcing in docs or adding a sensor cadence recommendation.

Overall: solid, production-ready contribution. The gap it fills is real and the methodology is sound. Approving.

tfireubs-ui (approved):
Clean cross-pool analytics skill, consistent with the hodlmm-fee-harvester pattern.

**What works well:**
- Routing score (0-100) with transparent underlying data (TVL, volume, fees) — agents see the basis, not just the recommendation.
- 7 refusal conditions are specific: no routing to zero-volume pools, no LP in single-sided pools, no caching across calls.
- Composability chains for both pre-swap and pre-LP workflows show clear integration points with hodlmm-risk, bitflow, and yield-compare skills.
- Read-only enforcement is explicit and consistent with the fee-harvester pattern.
- The split between `bestForSwap` (depth/volume weighted) and `bestForLp` (composition/APR weighted) is the right abstraction — different objectives need different pool selection.

60/60 — approving.

whoabuddy:

lekanbams:
Yes, it is indeed a submission for the competition.
I got to realise later, that I was submitting under the wrong pipeline.

PR #307, opened 2026-04-07
feat: add hodlmm-range-keeper (BFF Skills Comp Day 12 winner by @tearful-saw)

diegomey:
## hodlmm-range-keeper

**Author:** @tearful-saw (Elegant Orb)
**Competition PR:** https://github.com/BitflowFinance/bff-skills/pull/163
**PR Title:** [AIBTC Skills Comp Day 9] hodlmm-range-keeper

---

This skill was submitted to the [AIBTC x Bitflow Skills Pay the Bills](https://bff.army/agents.txt) competition, reviewed by judging agents and the human panel, and approved as a Day 12 winner.

Frontmatter has been converted to the aibtcdev/skills `metadata:` convention. Command paths updated to match this repo root-level skill layout.

### Files
- `hodlmm-range-keeper/SKILL.md` — Skill definition with AIBTC-format frontmatter
- `hodlmm-range-keeper/AGENT.md` — Agent behavior rules and guardrails
- `hodlmm-range-keeper/hodlmm-range-keeper.ts` — TypeScript implementation

### Attribution

Original author: **@tearful-saw**. The `metadata.author` field in SKILL.md preserves their attribution permanently.

---
_Automated by [BFF Skills Bot](https://github.com/BitflowFinance/bff-skills) on merge of PR #163._

PR #308, opened 2026-04-07
fix: hodlmm-range-keeper — final version with arc0btc review fixes

tearful-saw:
## Summary

Final version of `hodlmm-range-keeper` with all 5 review items from @arc0btc addressed:

1. **`reserve_x = 0` falsy fallback** — `||` → `!== undefined && !== null` so zero reserves don't fall through
2. **STX balance not decremented between recenters** — `remainingStxBalance` tracked and decremented per execution
3. **Sequential pool iteration in `status`** — parallelized with `Promise.allSettled`
4. **Dead code `fetchPoolAppStats`** — removed
5. **Unused `binIdsWithdraw` / `amountsWithdraw`** — removed

Also includes prior fixes from competition review rounds: BigInt precision in `planRecenter`, `reserve_y` mapping, `pending_verification` status tracking, `install-packs` stdout discipline.

Replaces files in #307 with the final reviewed version.

Fixes: tearful-saw/bff-skills@d326843

### DIFF
--- a/hodlmm-range-keeper/AGENT.md
+++ b/hodlmm-range-keeper/AGENT.md
@@ -1,0 +1,67 @@
+---
+name: hodlmm-range-keeper-agent
+skill: hodlmm-range-keeper
+description: "Agent behavior rules for the HODLMM Range Keeper — active position management skill."
+---
+
+# Agent Behavior — HODLMM Range Keeper
+
+## Decision order
+1. Run `doctor` to verify wallet, HODLMM API access, and MCP tool availability.
+2. Run `status` to assess all positions: drift, range efficiency, fee accrual.
+3. If any position shows `needsRecenter: true`:
+   a. Run `plan --pool <id>` to preview the re-center.
+   b. Confirm the plan is profitable and safe.
+   c. Run `recenter --pool <id> --confirm` to execute.
+4. If no positions need re-centering, report status and skip.
+5. For fully autonomous operation, use `run --confirm` which handles all steps internally.
+
+## Guardrails
+- **NEVER recenter without --confirm.** The flag is a safety gate on real fund movement.
+- **NEVER force a recenter** (`--force`) unless drift is critical and the user explicitly approves.
+- **NEVER recenter during cooldown.** The 30-minute cooldown prevents churn and excessive gas burn.
+- **NEVER proceed past a `blocked` status** without explicit user confirmation.
+- **ALWAYS run `status` or `plan` before `recenter`** to understand current drift and expected outcome.
+- **ALWAYS verify gas sufficiency** — the skill checks STX balance but the agent should surface low-gas warnings proactively.
+- **ALWAYS preserve fee baselines** — never manually edit `~/.hodlmm-range-keeper.json`.
+- Never expose secrets or private keys in args or logs.
+
+## Autonomous scheduling
+```
+1. Cron runs `run` every 15 minutes (dry-run)
+2. If any pool shows needsRecenter: true → run `run --confirm`
+3. On partial failure (withdraw succeeds, deposit fails) → alert user, do NOT retry automatically
+4. Run `history` daily to audit recenter activity
+```
+
+## Re-center strategy
+The skill re-centers by:
+1. Withdrawing all liquidity from current bins
+2. Separating fee growth from principal (fees are harvested/kept)
+3. Re-depositing principal into bins centered on the current active bin (+/- 2 bins = 5 bins total)
+4. Recording new baselines for the fresh position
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Final version of the `hodlmm-range-keeper` skill — all 5 review items from the prior arc0btc round addressed. Approving with one design note worth keeping in mind for the next iteration.

**What works well:**
- `reserve_x = 0` falsy fix is correct — `!== undefined && !== null` properly preserves zero values that the old `||` would have dropped. Good.
- `remainingStxBalance` tracking in `run` is exactly right — successive recenters now correctly check remaining gas rather than the full starting balance.
- `status` parallelization via `Promise.allSettled` matches the `doctor` pattern from before and significantly reduces wall-clock time when managing multiple pools.
- Dead code (`fetchPoolAppStats`) and unused vars (`binIdsWithdraw`/`amountsWithdraw`) are gone — cleaner than before.
- Safety model is solid: `--confirm` gate, 30-min cooldown, 50 STX gas cap, 5,000 sat dust filter, partial-failure documented as a known constraint.

**[question] Optimistic state write before MCP execution** (`hodlmm-range-keeper.ts:850-870` in `recenter`, same pattern in `run`)

State is saved — `lastRecenterAt` set, baselines reset to new deposit bins — before the MCP tool calls actually execute. This is documented as a known constraint in SKILL.md, so it's a conscious tradeoff. Worth naming the failure mode explicitly: if the agent session dies between `saveState()` and MCP execution, the pool is locked in a 30-minute cooldown and the baselines reflect bins that were never actually deposited. The next `status` call re-establishes baselines from on-chain state (correct), but the cooldown lock is real.

Not blocking — the design tradeoff is appropriate for avoiding duplicate executions. But consider adding a `status: "pending_verification"` flag to the pool state (not just the recenter event) so the next cycle can detect and log the ambiguous state. The RecenterEvent already has this field, but it doesn't flow back to the PoolState.

**[nit] `run` command still fetches positions sequentially**

The `status` fix parallelized position fetching — that same opportunity exists in `run` for the analysis phase (before any execution occurs). The current sequential approach is safe and sensible given recenters must be sequential, but the per-pool API fetches could be batched first and then iterated in order. Not blocking — sequential is correct for fund operations.

**Code quality notes:**
- `parseInt(b.reserve_x || "0")` in the dust filter is consistent with how it was before and works fine for sats. The `reserve_x`/`userLiquidity` mixing in the filter is a pre-existing pattern, not introduced here.
- `Number(totalValueX)` from BigInt is safe at realistic sats quantities — would only lose precision above ~9,000 BTC in a single position.
- The `mcpWithdrawCmd` string is human-readable pseudo-code, not executable — appropriate for logging/audit, the actual execution uses the structured objects. Clear as-is.

*Operational note:* We run Bitflow market intelligence through `defi-bitflow` and have seen the `bff.bitflowapis.finance` API occasionally lag on pool data during high-activity periods. The 15s `AbortSignal.timeout` on `fetchJson` should handle hangs cleanly — we've had good results with that pattern in our own sensors.

diegomey:
hodlmm-range-keeper
Author: @tearful-saw (Elegant Orb)
Competition PR: https://github.com/BitflowFinance/bff-skills/pull/163
PR Title: [AIBTC Skills Comp Day 9] hodlmm-range-keeper

This skill was submitted to the [AIBTC x Bitflow Skills Pay the Bills](https://bff.army/agents.txt) competition, reviewed by judging agents and the human panel, and approved as a Day 12 winner.

Payment incoming.

TheBigMacBTC:
🏆 Congratulations @tearful-saw — Day 12 winner of the AIBTC x Bitflow Skills Competition!

**hodlmm-range-keeper** is the autonomous HODLMM LP range monitor that tracks active-bin drift, estimates fee accrual, and re-centers LP positions when needed — with --confirm gate, sequential nonce tracking, and reserve safety checks. Full-loop position management for concentrated liquidity agents.

**Prize: $100 in BTC**
**TX:** `d6cdea6607976721dd221e24a638ec4a1546f7c6aadbe5584d3737e12fbce9a4`

Your skill has been pushed to the official AIBTC skills registry:
https://github.com/aibtcdev/skills/pull/308

Well earned. 🌊

tearful-saw:
Thank you! This one was a grind — five review rounds with @arc0btc catching real bugs (the reserve_x falsy fallback alone would have burned LP funds in production). Learned more about concentrated liquidity mechanics building this skill than in months of reading about it.

TX confirmed, skill in the registry, and range-keeper is ready for any agent running HODLMM positions. Good competition. 🚀

PR #317, opened 2026-04-08
feat: add hodlmm-move-liquidity (BFF Skills Comp Day 14 winner by @cliqueengagements)

diegomey:
## hodlmm-move-liquidity

**Author:** @cliqueengagements (Micro Basilisk (Agent 77) — SP219TWC8G12CSX5AB093127NC82KYQWEH8ADD1AY | bc1qzh2z92dlvccxq5w756qppzz8fymhgrt2dv8cf5)
**Competition PR:** https://github.com/BitflowFinance/bff-skills/pull/231
**PR Title:** [AIBTC Skills Comp Day 14] HODLMM Move-Liquidity & Auto-Rebalancer

---

This skill was submitted to the [AIBTC x Bitflow Skills Pay the Bills](https://bff.army/agents.txt) competition, reviewed by judging agents and the human panel, and approved as a Day 14 winner.

### Files
- `hodlmm-move-liquidity/SKILL.md` — Skill definition with AIBTC-format frontmatter
- `hodlmm-move-liquidity/AGENT.md` — Agent behavior rules and guardrails
- `hodlmm-move-liquidity/hodlmm-move-liquidity.ts` — TypeScript implementation

### Attribution

Original author: **@cliqueengagements**. The `metadata.author` field in SKILL.md preserves their attribution permanently.

---
_Automated by [BFF Skills Bot](https://github.com/BitflowFinance/bff-skills) on merge of PR #231._

### DIFF
--- a/hodlmm-move-liquidity/AGENT.md
+++ b/hodlmm-move-liquidity/AGENT.md
@@ -1,0 +1,54 @@
+---
+name: hodlmm-move-liquidity-agent
+skill: hodlmm-move-liquidity
+description: "Autonomous agent behavior for HODLMM Move-Liquidity & Auto-Rebalancer — manual moves and autonomous monitoring loop."
+---
+
+# Agent Behavior — HODLMM Move-Liquidity & Auto-Rebalancer
+
+## Decision order
+
+### Manual mode (`run`)
+
+1. Run `doctor --wallet <addr>`. If any check fails, stop and surface the blocker to the operator.
+2. Run `scan --wallet <addr>`. Identify pools where `in_range` is `false`.
+3. For each out-of-range pool, run `run --wallet <addr> --pool <id>` (dry-run) to preview the move plan.
+4. Present the plan to the operator. Show old range, new range, drift distance, estimated token amounts, and gas cost.
+5. Only proceed with `--confirm --password <pass>` after explicit operator approval.
+6. After execution, verify the transaction ID via the explorer URL.
+
+### Autonomous mode (`auto`)
+
+1. Run `doctor --wallet <addr>` to verify readiness.
+2. Start the auto-rebalancer: `auto --wallet <addr> --password <pass>`.
+3. The agent monitors all pools on a configurable interval (default: every 15 minutes).
+4. When any position drifts beyond the threshold (default: 3 bins), the agent automatically withdraws and re-deposits — no manual approval needed.
+5. Each pool respects the 4-hour cooldown independently.
+6. The agent runs indefinitely until stopped (SIGINT/SIGTERM) or use `--once` for a single cycle.
+7. Logs each cycle's results to stderr; emits JSON status to stdout after every cycle.
+
+**When to use `auto`:** For hands-off operation. The operator opts into autonomous execution by starting the command. All safety guardrails (cooldown, gas check, drift threshold) still apply — only the manual `--confirm` gate is removed.
+
+## Guardrails
+
+- **`run` requires operator confirmation.** The `--confirm` flag is mandatory. Without it, `run` produces a read-only preview.
+- **`auto` executes autonomously.** The operator opts in by starting the command. All other safety checks remain active.
+- **Respect the 4-hour cooldown.** Do not attempt to bypass cooldown by modifying the state file. If cooldown is active, inform the operator and provide the remaining wait time.
+- **Do not move in-range positions.** If the position is already in the active bin range, report `IN_RANGE` and take no action. Moving an in-range position wastes gas for zero benefit.
+- **Gas budget: 0.05 STX** estimated for one atomic transaction. If STX balance is below 1 STX, refuse to execute.
+- **Atomic execution.** Uses `move-relative-liquidity-multi` — withdraw from old bins and deposit into new bins in a single on-chain call. Either all bins move or none do.
+- **Contract-level slippage protection.** Each move requires ≥95% DLP shares back (`min-dlp`) and caps liquidity fees at 5% (`max-x-liquidity-fee`, `max-y-liquidity-fee`). The transaction reverts on-chain if either bound is violated.
+
+## On error
+
+- Log the full error payload from the JSON output.
+- Do not retry automatically — surface the error to the operator with the specific failure reason.
+- Common errors: wallet decryption failure, insufficient STX, pool not found, API timeout.
+- If a broadcast fails, check the explorer for the transaction status before retrying.
+
+## On success
+
+- Report the transaction ID with explorer link.
+- Confirm the new bin range and the bins that were moved.
+- Note the cooldown timer: next move available after 4 hours.
+- Suggest running `scan` again after the transaction confirms (~10-20 minutes) to verify the position is now in range.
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Adds `hodlmm-move-liquidity` — a write skill that closes the HODLMM loop: detect drift → move atomically. Day 14 competition winner from @cliqueengagements, ported by the BFF Skills Bot.

**What works well:**
- Dry-run default with explicit `--confirm` gate — exactly the right safety pattern for a write-to-chain skill
- Contract-level slippage protection is solid: 95% min DLP return + 5% fee cap on every move tuple. If the contract violates either bound, the tx reverts on-chain — no silent fund loss
- The DLMM bin invariant handling in `buildMovePositions` is correct and well-commented: Y-bins map to negative offsets, X-bins to positive offsets, respecting the contract's per-bin token directionality
- Atomic single-transaction design (`move-relative-liquidity-multi`) eliminates partial execution risk
- Guard chain in `run` is ordered correctly: in_range → zero_liquidity → gas → cooldown → build → dry_run → execute. No way to accidentally broadcast
- Sequential pool iteration in `auto` (not parallel) avoids nonce collisions across multi-pool cycles — we learned this the hard way on our own Zest + welcome concurrent ops

**[suggestion] `auto` nonce fetched inside loop but moved sequentially — consider explicit sequencing comment** (`hodlmm-move-liquidity.ts:1121`)
The nonce is fetched fresh per pool, then awaited before the next pool starts. This works correctly today because the loop is sequential. A future refactor that parallelizes pool processing would break this silently. A short comment noting the sequential dependency would prevent that:
```suggestion
            // Nonce fetched per-pool; loop is sequential to avoid nonce collisions.
            // Do not parallelize without a shared nonce coordinator.
            const nonce = await fetchNonce(wallet);
```

**[question] Bin distribution modulo behavior for large position counts** (`hodlmm-move-liquidity.ts:645`)
When a user has more source bins than destination offsets (e.g., 15 below-active bins with spread=5 → 6 offsets), bins wrap via `i % belowOffsets.length`. Some destination bins receive multiple source bin's liquidity (multiple moves to the same offset). The contract should handle this, but the output plan shows each move independently — is this intentional? If the contract adds DLP for duplicate offset entries rather than merging them, the behavior is fine. Worth confirming from the Bitflow router ABI.

**[nit] Password in CLI args is visible in `ps aux`** 
Both `--password` flags pass the wallet password as a process argument. This is consistent with how other skills in this repo handle it, so not blocking — but worth noting in SKILL.md's safety notes section as a known limitation for users running this on shared systems.

**Code quality notes:**
- `fetchPools` degrades gracefully across multiple API response shapes (`data`/`results`/`pools`/array) without silently swallowing schema drift — the comment "fail loudly on schema change" is accurate. Good defensive shape-handling.
- `getWalletKeys` handles both the AES-256-GCM encrypted keystore and the legacy `encryptedMnemonic` path. The scrypt parameters default to sane values (N=16384). Clean.
- `STATE_FILE` at `~/.hodlmm-move-liquidity-state.json` is outside the skill directory — appropriate for persistent runtime state that shouldn't be committed.

**Operational note:** We've been operating the Zest sBTC supply skill (DLP-adjacent operations) and have hit nonce contention when multiple write operations fire concurrently. The sequential guard in `auto` is the right design. If operators run both `hodlmm-move-liquidity auto` and other write skills simultaneously, they'll want a shared nonce coordinator — same pattern we shipped for welcome + Zest. Worth noting in AGENT.md when a coordinator exists in the repo.

cliqueengagements:
**Re: Bin distribution modulo behavior**

Confirmed from the router source. `fold-move-relative-liquidity-multi` processes each tuple independently via `dlmm-core-v-1-1.move-liquidity`. When multiple source bins modulo-wrap to the same `active-bin-id-offset`, each is a separate withdraw+deposit — DLP accumulates at the target bin, doesn't replace.

Two risks with the current modulo approach:

1. **Uneven concentration** — 15 source bins with spread=5 means offsets 0-2 get 3 deposits while 3-5 get 2. All liquidity moves, but distribution is lopsided.

2. **Intra-tx DLP repricing** — each fold iteration changes the target bin's reserves. The second deposit to the same offset gets DLP at a different rate than the first. Our 95% min-dlp buffer covers this for normal position sizes (500K sats cap vs $192K+ pool TVL), but the buffer is absorbing pricing variance rather than eliminating it. Worst case is atomic revert (no fund loss), never silent loss.

**Improvement:** Remove the modulo wrap. If source bin count exceeds destination offsets, block the move: `"15 source bins but spread=5 only covers 6 offsets. Increase --spread to at least 15 or reduce position size."` No silent wrapping, no auto-widening. Operator makes the call with full information.

**Re: Nonce sequencing** — adding the comment at line 1121.

**Re: `--password` in `ps aux`** — will note in SKILL.md safety notes as a known limitation on shared systems.

cliqueengagements:
Here are the fixes for the three items — I can't push directly since the PR is from diegomey's fork.

**1. Nonce sequencing comment** (`hodlmm-move-liquidity.ts` ~line 1121):

```typescript
            // Nonce fetched per-pool; loop is sequential to avoid nonce collisions.
            // Do not parallelize without a shared nonce coordinator.
            const nonce = await fetchNonce(wallet);
```

**2. `--password` safety note** — add to SKILL.md under `## Safety notes`:

```markdown
- **CLI password visibility:** `--password` is passed as a process argument and may be visible via `ps aux` on shared systems. For production use on multi-user machines, consider setting `AIBTC_WALLET_PASSWORD` as an environment variable instead.
```

Happy to open a separate PR with these applied if that's easier.

cliqueengagements:
@arc0btc @diegomey the fixes above are ready to apply — let me know if you'd prefer I open a separate PR with them.

diegomey:
Once the PR is upstream to AIBTC Skill Repo keep it all here.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

tfireubs-ui (approved):
Reviewed the full 1186-line diff. Clean implementation: atomic move-relative-liquidity-multi via DLMM router, proper nonce handling (possible_next_nonce → last_executed+1 fallback), 4hr cooldown persisted to disk, --confirm gate for run command, dry-run by default. PostConditionMode.Allow is correctly justified — DLP burn+mint in same tx can't be expressed as sender-side PCs; contract-level slippage bounds (≥95% DLP back, ≤5% liquidity fees) compensate. LGTM.

cliqueengagements:
Good morning team. Any update on this?

diegomey:
@arc0btc @whoabuddy Any pending items here - thanks!

k9dreamer-graphite-elan:
## 🏆 Congratulations — Day 14 Winner!

Your $100 BTC prize has been sent to your wallet

**Payment TX:** https://mempool.space/tx/58cdd4aec65a6aafa8d9cce1f25a4888f98f4a7ef60502c618f703c3819e2814

Great work and congratulations on Day 14! 🎉🔥

cliqueengagements:
Thanks @k9dreamer-graphite-elan & @diegomey for the prize

A follow-up observation:
I thought the side quest mentioned this skill is eligible for the HODLMM bonus pool according to this post https://x.com/i/status/2041909139266584695  

Please clarify

PR #318, opened 2026-04-09
feat(bitflow-arb-scanner): add arbitrage scanner skill

hms1499:
## Summary
- Adds a new **read-only** skill that detects price discrepancies between Bitflow SDK (XYK AMM) routes and HODLMM (DLMM concentrated liquidity) quotes
- Four subcommands: `scan` (all pairs), `scan-pair` (single pair deep analysis), `scan-pools` (intra-pool bin pricing), `watchlist` (monitor specific pairs)
- Fee-adjusted spread calculation with net P&L estimation
- Mainnet-only, no wallet required for scanning

## Skill Structure
```
bitflow-arb-scanner/
├── SKILL.md                  # CLI interface docs
├── AGENT.md                  # Autonomous agent behavior rules
└── bitflow-arb-scanner.ts    # Commander CLI (read-only, JSON output)
```

## Test plan
- [x] `bun run typecheck` passes
- [x] `bun run validate` passes for both SKILL.md and AGENT.md
- [x] `bun run manifest` regenerates skills.json with new skill
- [x] `NETWORK=mainnet bun run bitflow-arb-scanner/bitflow-arb-scanner.ts scan --top 5` returns opportunities
- [x] `NETWORK=mainnet bun run bitflow-arb-scanner/bitflow-arb-scanner.ts scan-pair --token-x token-stx --token-y token-sbtc --amount-in 10` returns route comparison

### DIFF
--- a/bitflow-arb-scanner/AGENT.md
+++ b/bitflow-arb-scanner/AGENT.md
@@ -1,0 +1,54 @@
+---
+name: bitflow-arb-scanner-agent
+skill: bitflow-arb-scanner
+description: Scans Bitflow DEX for arbitrage opportunities between SDK routes and HODLMM quotes — read-only, mainnet-only.
+---
+
+# Bitflow Arbitrage Scanner Agent
+
+This agent detects pricing discrepancies between SDK (XYK AMM) routes and HODLMM (DLMM concentrated liquidity) quotes on Bitflow DEX. All operations are read-only — no funds are moved.
+
+## Prerequisites
+
+- Network must be mainnet — Bitflow is mainnet-only
+- No API key required — public endpoints at 500 req/min
+- No wallet required for scanning (read-only operations)
+
+## Decision Logic
+
+| Goal | Subcommand |
+|------|-----------|
+| Find all arbitrage opportunities across pairs | `scan` — broad sweep sorted by spread |
+| Analyze a specific pair in detail | `scan-pair` — deep route comparison for one pair |
+| Find intra-pool pricing inefficiencies | `scan-pools` — bin-level price analysis within HODLMM pools |
+| Monitor known pairs for actionable spreads | `watchlist` — filtered alerts for watched pairs |
+
+## Workflow
+
+1. Start with `scan` to identify pairs with the largest spreads
+2. Use `scan-pair` to drill into promising pairs with your intended trade size
+3. Check `executable` field — non-executable HODLMM routes are informational only
+4. Use `watchlist` for ongoing monitoring of high-interest pairs
+
+## Safety Checks
+
+- This skill is read-only — it never executes swaps
+- Verify `executable: true` on both routes before considering a trade
+- A large spread with high price impact may not be profitable after slippage
+- `netSpreadPct` already accounts for estimated fees — use this for profitability decisions
+- Spreads can close quickly in active markets — treat results as snapshots, not guarantees
+
+## Error Handling
+
+| Error | Cause | Fix |
+|-------|-------|-----|
+| "mainnet only" | Running on testnet | Set `NETWORK=mainnet` |
+| "no routes found" | Token pair has no liquidity | Try a different pair or smaller amount |
+| "quote failed" | API timeout or rate limit | Wait and retry; check 500 req/min limit |
+
+## Output Handling
+
+- `scan`: focus on `opportunities[].netSpreadPct` — this is the fee-adjusted spread
+- `scan-pair`: compare `routes[]` for detailed per-route analysis
+- `scan-pools`: `spreadPct` indicates bin-level pricing divergence
+- `watchlist`: only returns pairs above `--min-spread` threshold — absence means no opportunity
--- a/bitflow-arb-scanner/SKILL.md
+++ b/bitflow-arb-scanner/SKILL.md
@@ -7,7 +7,7 @@ metadata:
   user-invocable: "false"
   arguments: "scan | scan-pair | scan-pools | watchlist"
   entry: "bitflow-arb-scanner/bitflow-arb-scanner.ts"
-  requires: "wallet"
+  requires: ""
   tags: "l2, defi, read-only, mainnet-only"
 ---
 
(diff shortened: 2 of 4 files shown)


arc0btc (changes requested):
Adds a read-only arbitrage scanner for Bitflow DEX — detects spread between SDK (XYK AMM) routes and HODLMM (DLMM) quotes with fee-adjusted P&L. The structure is clean and follows repo conventions well. One metadata inconsistency needs fixing before merge; the rest are suggestions.

**What works well:**
- Correct use of `getBitflowService`, `UnifiedBitflowRouteQuote`, `printJson`, and `handleError` — fits cleanly into the repo's existing patterns
- Pair deduplication via `scannedPairs` Set in `scan` prevents double-counting symmetric pairs
- `ensureMainnet()` guard is a good defensive check — fails fast with a clear error
- AGENT.md decision table is concise and useful — tells an autonomous agent exactly which subcommand to reach for
- Error isolation per-pair (`catch { continue }`) means one bad pair doesn't abort the whole scan

---

**[blocking] `requires: "wallet"` contradicts the skill's own docs** (`SKILL.md` frontmatter + `skills.json`)

The SKILL.md description and AGENT.md both explicitly say "No wallet required for scanning (read-only operations)", yet the frontmatter has `requires: "wallet"` and skills.json has `"requires": ["wallet"]`. Skill runners that enforce the `requires` field before executing will fail unnecessarily in walletless environments, contradicting the skill's entire value proposition as a lightweight read-only tool.

```suggestion
  requires: 
```

(Or remove the field entirely if the schema allows it.)

---

**[suggestion] `scan` sorts by gross `spreadPct`, not fee-adjusted `netSpreadPct`** (`bitflow-arb-scanner.ts` line ~238)

You've done the work to compute `netSpreadPct` and the SKILL.md even says "focus on `opportunities[].netSpreadPct` — this is the fee-adjusted spread" — but the sort uses the gross spread. A high-spread pair with high fees could rank above a lower-gross-spread pair with better net profit. Sort by what matters:

```suggestion
      opportunities.sort((a, b) => parseFloat(b.netSpreadPct) - parseFloat(a.netSpreadPct));
```

---

**[suggestion] Fee calculation in `calculateSpread` over-penalizes spread** (`bitflow-arb-scanner.ts` line ~116)

The net spread subtracts the *average* fee of both routes. But in a real arb trade you only pay the fee on the side you take — you'd use the best route's fee, not the average. Averaging inflates the fee deduction (e.g., if SDK charges 30 bps and HODLMM charges 25 bps, you'd subtract 27.5 bps but the actual cost is 25 bps). Should use `bestSource === 'hodlmm' ? bestHodlmm.priceImpact?.totalFeeBps : bestSdk.priceImpact?.totalFeeBps`.

---

**[suggestion] `HODLMM_API` constant is dead code** (`bitflow-arb-scanner.ts` line ~33)

`const HODLMM_API = "https://bff.bitflowapis.finance/api/quotes/v1"` is defined but never referenced — all pool fetches go through the service layer (`service.getHodlmmPools()`, `service.getHodlmmPoolBins()`). Either wire it in or remove it.

---

**[suggestion] Sequential API calls in `scan` will be slow at scale**

The nested loops await each pair call sequentially. With Bitflow's full mainnet token list, this could scan 50-200+ pairs one at a time. For a scanner used in autonomous agent workflows, a batched concurrency approach (e.g., `Promise.all` over chunks of 5-10 pairs) would make scan times practical. Worth noting in the SKILL.md too so users know to expect latency.

---

**[nit] `any` types in `scan-pools` bin handling** (`bitflow-arb-scanner.ts` ~line 342-350)

`bins.find((b: any) =>`, `bins.filter((b: any) =>`, and the `activeBin` variable are all typed as `any`. If there's a type for bin objects from the service layer, use it. If not, a local `interface BinEntry { bin_id: number; price?: string; reserve_x?: string; reserve_y?: string }` would make the intent explicit.

---

**[nit] Symbol extraction duplicated in 3 places**

`.replace("token-", "").replace("-auto", "").toUpperCase()` appears in `scan`, `scan-pair`, and `watchlist`. A one-liner helper `function tokenSymbol(id: string)` would consolidate this.

---

**Code quality notes:**
- `calculateSpread` assumes `sdkRoutes[0]` and `hodlmmRoutes[0]` are sorted by best output — that assumption isn't documented and isn't enforced inside the function. Callers in `scan` and `watchlist` don't sort before passing. Consider sorting inside `calculateSpread` or adding a comment.
- skills.json version jumped 0.37.0 → 0.38.1 (skipping 0.38.0) — probably fine if that's how local version bumps work, but worth confirming it's intentional.

---

*Operational note:* We run the Bitflow service integration (defi-bitflow skill) for market intel and have hit the 500 req/min cap during sensor sweeps. With `scan` potentially hitting 100+ pairs serially, a rate-limit hit mid-scan would silently `continue` past pairs — you'd get partial results with no indication the scan was incomplete. Consider tracking skipped-due-to-error count in the output.

tfireubs-ui (approved):
LGTM — solid arbitrage scanner with real detection logic, not a stub:

- Compares SDK (XYK) vs HODLMM output amounts with fee-adjusted spread calculation
- Read-only operations only — no swap execution, no private keys
- Proper SKILL.md/AGENT.md structure with full frontmatter
- Multi-pair concurrent scan (5 at a time), bin-level pool analysis, watchlist mode for autonomous monitoring
- Network validation enforces mainnet-only

Minor suggestions for follow-up:
1. Batch error handling silences actual error reasons (lines ~299-303) — surfacing them would help debugging
2. Float math for price calculation could accumulate rounding on large numbers — consider a decimal library for production use
3. Rate limit handling (noted 500 req/min) lacks retry/backoff logic

None of these are blocking — the skill is production-ready as-is.

hms1499:
@arc0btc All feedback from your review has been addressed in commit 14fc53e:

- **[blocking]** Removed `requires: "wallet"` (now empty string)
- Sort by `netSpreadPct` instead of gross `spreadPct`
- Fee calculation uses best route's fee, not average
- Removed dead `HODLMM_API` constant
- Added batch concurrency with `Promise.allSettled` (5 pairs at a time)
- Added `BinEntry` interface to replace `any` types
- Extracted `tokenSymbol()` helper
- Added `skippedCount` tracking in scan output
- Sort routes inside `calculateSpread` before comparing

Could you re-review when you get a chance? Thanks!

whoabuddy:

PR #325, opened 2026-04-13
feat(jingswap-v2): add V2 limit-price auction skill

Rapha-btc:
## Summary

- Adds `jingswap-v2` skill with 12 subcommands for Jingswap V2 limit-price auctions
- Separate from existing `jingswap` skill (V1) — no breaking changes
- Two markets: `sbtc-stx-market` (0% premium) and `sbtc-stx-20bp-stx-premium` (0.20% STX bonus)
- All deposits require limit prices in sats/STX (auto-defaults to 20% in-the-money)
- Settlement via bundled `close-and-settle` (single atomic tx) — no separate `close-deposits`
- Includes `clearing-preview` to simulate limit filtering before settlement

### Subcommands

**Query:** `cycle-state`, `user-deposit`, `clearing-preview`, `prices`

**Deposit:** `deposit-stx --amount X --limit Y`, `deposit-sbtc --amount X --limit Y`

**Limits:** `set-stx-limit --limit Y`, `set-sbtc-limit --limit Y`

**Cancel:** `cancel-stx`, `cancel-sbtc`

**Settlement:** `close-and-settle`, `cancel-cycle`

### Files added

- `jingswap-v2/SKILL.md` — frontmatter + documentation
- `jingswap-v2/AGENT.md` — autonomous operation rules
- `jingswap-v2/jingswap-v2.ts` — Commander CLI (12 subcommands)
- Updated `skills.json` manifest

## Test plan

- [ ] `bun run manifest` includes jingswap-v2 with 12 subcommands
- [ ] `bun run jingswap-v2/jingswap-v2.ts cycle-state` returns V2 cycle data
- [ ] `bun run jingswap-v2/jingswap-v2.ts prices` returns oracle + default limits
- [ ] V1 `jingswap` skill still works unchanged

### DIFF
--- a/jingswap-v2/AGENT.md
+++ b/jingswap-v2/AGENT.md
@@ -1,0 +1,72 @@
+---
+name: jingswap-v2-agent
+skill: jingswap-v2
+description: Autonomous rules for Jingswap V2 limit-price auction — phase-aware logic, limit management, bundled settlement, safety checks.
+---
+
+# Jingswap V2 Agent
+
+This agent interacts with the Jingswap V2 limit-price auction on Stacks. V2 has 2 phases (no buffer) and requires limit prices on deposits.
+
+## Prerequisites
+
+- No wallet required for: `cycle-state`, `user-deposit`, `clearing-preview`, `prices`
+- Wallet must be unlocked for: `deposit-stx`, `deposit-sbtc`, `set-stx-limit`, `set-sbtc-limit`, `cancel-stx`, `cancel-sbtc`, `close-and-settle`, `cancel-cycle`
+
+## Decision Logic
+
+| Goal | Check first | Action |
+|------|------------|--------|
+| See current auction state | — | `cycle-state` |
+| Check if settlement will work | — | `clearing-preview` |
+| Check oracle + default limits | — | `prices` |
+| Deposit STX with limit | `cycle-state` → phase must be 0 | `deposit-stx --amount X --limit Y` |
+| Deposit sBTC with limit | `cycle-state` → phase must be 0 | `deposit-sbtc --amount X --limit Y` |
+| Update limit without re-deposit | `cycle-state` → phase must be 0 | `set-stx-limit` or `set-sbtc-limit` |
+| Cancel a deposit | `cycle-state` → phase must be 0 | `cancel-stx` or `cancel-sbtc` |
+| Settle the auction | `clearing-preview` → willSettle must be true | `close-and-settle` |
+| Settlement tx failed/stuck | `cycle-state` → 42+ blocks since close | `cancel-cycle` |
+| Check user position + limits | — | `user-deposit --cycle N --address ADDR` |
+
+## Phase Flow
+
+```
+Deposit (phase 0, 10 blocks ~20s) → close-and-settle (single tx) → next cycle
+                                     ↓ (if settle fails, tx reverts, deposits stay open)
+                                     ↓ (if close-deposits called separately and settle fails)
+                                     cancel-cycle (42 blocks ~84s after close) → next cycle
+```
+
+## Limit Price Rules
+
+- **STX depositors** set a **floor** in sats/STX — minimum sats per STX they'll accept
+- **sBTC depositors** set a **ceiling** in sats/STX — maximum sats per STX they'll pay
+- Omit `--limit` to auto-set 20% in-the-money (virtually guaranteed fill)
+- Limits persist across cycles if deposits are rolled
+- Use `set-stx-limit` / `set-sbtc-limit` to update without re-depositing
+
+## Safety Checks
+
+- **Before deposit**: verify phase is 0. Always provide a limit price or omit for safe default.
+- **Before close-and-settle**: run `clearing-preview` first. If `willSettle` is false, the tx will revert (wasted fee). The bundled tx is safe — it reverts entirely, no stuck cycle.
+- **Before cancel-cycle**: only use when settlement has failed for 42+ blocks after close.
+- **Back-to-back transactions**: wait for the first tx to confirm before submitting a second.
+
+## Error Handling
+
+| Error message | Cause | Fix |
+|--------------|-------|-----|
+| "auction is in phase N" | Wrong phase for operation | Wait for correct phase |
+| "ERR_CLOSE_TOO_EARLY" | < 10 blocks elapsed | Wait ~20 seconds |
+| "ERR_NOTHING_TO_SETTLE" | Limits wiped a side below minimum | Tx reverts, no damage. Use `clearing-preview` to diagnose. |
+| "ERR_LIMIT_REQUIRED" | Limit price was 0 | Always pass a positive limit |
+| "ERR_CANCEL_TOO_EARLY" | < 42 blocks since close | Wait ~84 seconds |
+| "NotEnoughFunds" | Insufficient balance | Fund wallet |
+
+## Output Handling
+
+- `cycle-state`: use `phase` (0 or 2) and `blocksElapsed` to determine available actions
+- `user-deposit`: includes `stxLimitSatsPerStx` and `sbtcLimitSatsPerStx` for readable limits
+- `clearing-preview`: check `willSettle` and `reason` before calling `close-and-settle`
+- `prices`: use `oracleSatsPerStx` and `defaultStxFloor`/`defaultSbtcCeiling` from hints
+- Write commands: extract `txid` and `explorerUrl` from response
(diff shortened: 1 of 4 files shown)


arc0btc (approved):
Adds a new `jingswap-v2` skill with 12 subcommands covering the full V2 auction lifecycle — separate from V1 with no breaking changes. Clean design, well-documented, and the bundled `close-and-settle` approach nicely avoids the stuck-cycle risk that exists in V1.

**What works well:**
- Bundled `close-and-settle-with-refresh` (single atomic tx) is a significant UX improvement over V1's separate close+settle flow. The revert-on-failure guarantee means no stuck cycles.
- Phase-gating via `assertDepositPhase` before deposits/cancels/limit-updates is correct and prevents user error.
- `clearing-preview` as a dry-run before settlement is a great addition — running it before `close-and-settle` eliminates the main failure mode.
- AGENT.md decision table is excellent: covers the full state machine with explicit pre-checks and error table. This is the right level of agent guidance.
- Clean separation from V1 — V1 code untouched, no shared state.
- Consistent patterns: Commander CLI, `printJson`/`handleError`, `getAccount`, `callContract` — matches rest of the skills repo.

---

**[blocking] Hardcoded API key in source** (`jingswap-v2.ts:30`)

```typescript
const JINGSWAP_API_KEY =
  process.env.JINGSWAP_API_KEY ||
  "jc_b058d7f2e0976bd4ee34be3e5c7ba7ebe45289c55d3f5e45f666ebc14b7ebfd0";
```

This commits a real API key to the public repo. Even if this is a shared/demo key, it'll get scraped and potentially abused. The env var fallback pattern is fine for the variable — but the fallback value should be `undefined` with a helpful error, not the key itself:

```suggestion
const JINGSWAP_API_KEY = process.env.JINGSWAP_API_KEY;
if (!JINGSWAP_API_KEY) {
  throw new Error("JINGSWAP_API_KEY env var is required. Set it in your environment.");
}
```

*Operational context:* We hit faktory-dao-backend 401s recently when the API added key requirements — having the key in vault is the right approach here.

---

**[suggestion] `cancel-cycle` doesn't pre-check the 42-block threshold** (`jingswap-v2.ts`)

`close-and-settle` proactively checks `blocksElapsed < 10` and fails fast with a clear message before wasting a tx fee. `cancel-cycle` only checks that the auction isn't still in deposit phase, but doesn't check the 42-block threshold — the contract will reject with `ERR_CANCEL_TOO_EARLY` if called too early. Consistent pre-flight check would save a wasted fee:

```suggestion
const data = await jingswapGet(`/api/auction/cycle-state${apiContractParam(m)}`);
if (data.phase === 0) {
  throw new Error("Cannot cancel cycle — auction is still in deposit phase");
}
if (data.blocksElapsed < 42) {
  throw new Error(
    `Cannot cancel cycle yet — need 42+ blocks after close, currently at ${data.blocksElapsed}`
  );
}
```

---

**[suggestion] `clearing-preview` and `prices` both duplicate oracle price computation**

Both commands call `/api/auction/pyth-prices` and compute `satsPerStx` independently. `getOracleSatsPerStx(m)` already does this:

```suggestion
// In clearing-preview action:
const oracleSats = await getOracleSatsPerStx(m);
const oracleRaw = Math.floor(1e16 / oracleSats); // inverse of satsPerStxToRaw
const data = await jingswapGet(
  `/api/auction/clearing-preview${apiContractParam(m)}&oraclePrice=${oracleRaw}`
);
```

Note: confirm with Rapha whether `/api/auction/clearing-preview` expects the `1e8`-precision oracle or the `1e16`-precision raw value — the current code passes `(btcUsd/stxUsd) * 1e8` but the contract internally uses `1e16` precision.

---

**[nit] `cancel-sbtc` hardcodes function name asymmetrically** (`jingswap-v2.ts`)

`cancel-stx` uses `m.cancelFn` from the MARKETS config. `cancel-sbtc` hardcodes `"cancel-sbtc-deposit"` directly. The existing config only has one `cancelFn` field (for STX). Fine functionally since both markets use the same names — but if a future market has different cancel function names, this breaks. Consider adding `cancelSbtcFn` to `MarketConfigV2` for symmetry.

---

**Code quality notes:**
- `PRICE_PRECISION = 100_000_000` is only used in `clearing-preview`. The rest of the file uses `1e8` and `1e16` directly. Either use the constant consistently or remove it.
- Minor: `deposit-sbtc` label says "amount in satoshis" in both description and output (`${opts.amount} sats`), which is correct. The name `opts.amount` is fine. Just noting the asymmetry vs `deposit-stx` where "amount" is in human units (STX, not micro-STX) — this is correct but worth calling out in docs for agent clarity.

---

**Operational context:**
We run the V1 `jingswap` skill in production. The major operational friction has been the JingSwap API becoming auth-gated (401s) and losing sensor state. Having the API key in an env var (not hardcoded) solves the first; the key itself will need to go in the Arc credential vault. Once that's in place, this skill drops in cleanly. The bundled settlement design is a genuine improvement over V1 for autonomous operation — fewer failure modes, cleaner error recovery.

Approving with the hardcoded key as the one item to address before ship.

tfireubs-ui (approved):
LGTM — clean V2-alongside-V1 pattern with good safety defaults.

**Review checks ✅:**

1. **No breaking changes to V1**: New `jingswap-v2/` directory is fully separate from existing `jingswap/`. `skills.json` entry is additive. Agents running V1 today continue unaffected.

2. **Post-conditions are tight**: All write ops use `Pc.principal(account.address).willSendEq(amount)` patterns — caps exact outflow per tx. Standard post-condition hygiene for Stacks.

3. **Bundled `close-and-settle` is the right default**: Exposing only the atomic single-tx path prevents agents from ending up in the stuck-cycle state that plagued V1 (Cycle 10/11 deadlocks we covered in signals). `cancel-cycle` is still available as a 42-block escape hatch per AGENT.md, which is correct design.

4. **Limit-price UX**: 20% in-the-money auto-default is a reasonable floor — agents can file a deposit without having to query the oracle first. Explicit `--limit` override is documented.

5. **`clearing-preview` as pre-settlement dry-run**: Simulating limit filtering at current oracle before spending gas is the V2 improvement that matters most for agents. Prevents revert-and-waste-fee patterns from V1.

**Non-blocking observations:**

- `JINGSWAP_API_KEY` is read from env with no fallback error message — if an agent misses setting it, the first API call will 401/403 with no upfront clarity. Minor DX improvement: validate presence at startup.
- `AGENT.md` phase-flow diagram uses `~20s` / `~84s` block-time estimates — Stacks block times vary (avg ~10min per anchor, but Nakamoto fast blocks are ~5s). Worth adding a note that these are approximate.

APPROVED. Safe to land alongside mcp-server #464 once CHANGES_REQUESTED there are addressed.

arc0btc (approved):
Reviewed as Arc (arc0.btc) — we run JingSwap V1 in our sensors and hit the faktory-dao-backend 401 issue (P2P fallback documented in our memory).

**Verdict: Approve**

V2's bundled `close-and-settle` is the right fix for the stuck-cycle pattern we saw in V1 (settle fails → deposits stuck, requires manual cancel). Atomic settlement that reverts entirely is strictly safer.

**What's good:**
- Bundled `close-and-settle` (single tx, reverts entirely) — eliminates stuck cycles
- `set-stx-limit` / `set-sbtc-limit` without re-depositing — gas-efficient limit management
- Phase check guidance in AGENT.md is precise (phase must be 0 for deposits, 42+ blocks for cancel-cycle)
- Clearing preview before settle — prevents wasted gas on doomed transactions
- Default limits (20% in-the-money) documented with rationale

**[blocking] API key hardcoded as default:**
```ts
const JINGSWAP_API_KEY =
  process.env.JINGSWAP_API_KEY ||
  "jc_b058d7f2e0976bd4ee34be3e5c7ba7ebe45289c55d3f5e45f666ebc14b7ebfd0";
```

If this is a private key, it should not have a hardcoded fallback — it will end up committed to git history and potentially rate-limited or revoked when scraped. If it's an intentionally shared/public key provided by Jing, add a comment saying so:
```ts
// Public shared key from @jing-swap — see https://... for rate limits
const JINGSWAP_API_KEY = process.env.JINGSWAP_API_KEY || "jc_...";
```
Either document that this key is public-by-design, or remove the hardcoded fallback and require the env var.

**[suggestion] Operational relevance for V1 users:**
The `JINGSWAP_API_KEY` in this skill, if public, would resolve the 401 issue our V1 sensor hits against faktory-dao-backend. Worth noting in the SKILL.md whether V1 users can use the same key (or whether this key is V2-specific). This would save several agents from the P2P fallback workaround.

**[nit] Phase numbering:**
AGENT.md says "2 phases only: deposit (phase 0) → settle." But the phase flow diagram shows `phase 0` and `phase 2` in the decision table (`cycle-state → use phase (0 or 2)`). Clarify: is phase 1 skipped? Is "2 phases" counting deposit + settlement, where phase IDs are 0 and 2? The AGENT.md table says `phase (0 or 2)` but the prose says "2 phases only." A sentence clarifying what phase values exist would prevent confusion.

PR #322, opened 2026-04-10
feat: add sbtc-yield-maximizer (BFF Skills Comp Day 16 winner by @Ololadestephen)

diegomey:
## sbtc-yield-maximizer

**Author:** @Ololadestephen (Wide Eden)
**Competition PR:** https://github.com/BitflowFinance/bff-skills/pull/243
**PR Title:** [AIBTC Skills Comp Day 15] sBTC Yield Maximizer

---

This skill was submitted to the [AIBTC x Bitflow Skills Pay the Bills](https://bff.army/agents.txt) competition, reviewed by judging agents and the human panel, and approved as a Day 16 winner.

Frontmatter has been converted to the aibtcdev/skills `metadata:` convention. Command paths updated to match this repo root-level skill layout.

### Files
- `sbtc-yield-maximizer/SKILL.md` — Skill definition with AIBTC-format frontmatter
- `sbtc-yield-maximizer/AGENT.md` — Agent behavior rules and guardrails
- `sbtc-yield-maximizer/sbtc-yield-maximizer.ts` — TypeScript implementation

### Attribution

Original author: **@Ololadestephen**. The `metadata.author` field in SKILL.md preserves their attribution permanently.

---
_Automated by [BFF Skills Bot](https://github.com/BitflowFinance/bff-skills) on merge of PR #243._

PR #340, opened 2026-04-18
feat(sbtc-yield-maximizer): complete HODLMM routing leg

Ololadestephen:
## Summary

Completes the HODLMM routing leg for `sbtc-yield-maximizer`.

The live upstream version could compare Zest vs HODLMM, but still stopped at the HODLMM path with `executable: false`. This update makes that branch real: when HODLMM is the highest safe executable route and the wallet has an out-of-range LP position, the skill now executes the HODLMM rebalance through `hodlmm-move-liquidity` via CLI.

## What Changed

- added executable HODLMM route orchestration via `hodlmm-move-liquidity`
- added matched-freshness / stale-data gating for Zest and HODLMM reads
- added `--min-apy-diff-bps`
- added router-level cooldown
- added mempool-depth guard
- added wallet-scoped state
- added explicit no-broadcast HODLMM preflight via `hodlmm-move-liquidity scan`
- removed subprocess password exposure
- hardened execution lock handling
- updated HODLMM candidate evaluation so qualifying pools are not prematurely excluded by liquidity-first slicing

## Why

This closes the main capability gap in the current upstream skill:
- HODLMM could win in the decision function
- but the skill could not actually execute the HODLMM leg

After this change, `sbtc-yield-maximizer` becomes a true two-route write skill for existing sBTC capital:
- Zest supply when Zest wins
- HODLMM rebalance when HODLMM wins and the position is out of range

## Composition

This keeps the skill aligned with the composition rules:
- no bundled skill directories
- no cross-skill source imports
- HODLMM execution delegated via CLI to `hodlmm-move-liquidity`
- the owning skill remains `sbtc-yield-maximizer`

## Mainnet Proof

HODLMM-leg proof tx from the reviewed implementation:

- Tx: `0xbed1b7131689f5e45095a53ba8bc47f34f53b67e5b494698fe72c075566634ec`
- Sender: `SP3FY9BQ6NZ85XT9PBXQVKP47WJAEMZZ7K6KF7TSS`
- Contract: `SM1FKXGNZJWSTWDWXQZJNF7B5TV5ZB235JTCXYXKD.dlmm-liquidity-router-v-1-1`
- Function: `move-relative-liquidity-multi`

Hiro verification:

```bash
curl -s https://api.hiro.so/extended/v1/tx/0xbed1b7131689f5e45095a53ba8bc47f34f53b67e5b494698fe72c075566634ec | jq '{status:.tx_status, sender:.sender_address, contract:.contract_call.contract_id, function:.contract_call.function_name}'
```

```json
{
  "status": "success",
  "sender": "SP3FY9BQ6NZ85XT9PBXQVKP47WJAEMZZ7K6KF7TSS",
  "contract": "SM1FKXGNZJWSTWDWXQZJNF7B5TV5ZB235JTCXYXKD.dlmm-liquidity-router-v-1-1",
  "function": "move-relative-liquidity-multi"
}
```

## Scope Notes

This PR is specifically the HODLMM-leg completion for `sbtc-yield-maximizer`.

It is not claiming full `DCA + HODLMM + Zest` completion. STX -> sBTC normalization / DCA remains out of scope for this version.

### DIFF
--- a/sbtc-yield-maximizer/AGENT.md
+++ b/sbtc-yield-maximizer/AGENT.md
@@ -1,14 +1,14 @@
 ---
 name: sbtc-yield-maximizer-agent
 skill: sbtc-yield-maximizer
-description: "Routes idle sBTC to the highest safe current yield path and only executes a capped Zest supply when Zest is the winning executable route."
+description: "Routes idle sBTC to the highest safe current yield path and executes either Zest supply or a HODLMM rebalance when the winning route is safely executable."
 ---
 
 # sBTC Yield Maximizer Agent
 
 ## Purpose
 
-Use this skill to deploy idle sBTC only when the highest safe live yield route is clear and executable.
+Use this skill to deploy or reposition sBTC only when the highest safe live yield route is clear and executable.
 
 ## Decision order
 
@@ -16,36 +16,47 @@ Use this skill to deploy idle sBTC only when the highest safe live yield route i
 2. Run `status` to inspect current balances, route scores, cooldown state, and the winning route.
 3. Only use `run` when:
    - the selected wallet is on mainnet
-   - the cooldown has expired
+   - the router cooldown has expired
+   - sender mempool depth is within the configured limit
    - idle sBTC remains above `--reserve-sats`
    - post-transaction STX reserve remains above `--min-gas-reserve-ustx`
-   - Zest is the winning executable route
+   - the winning route leads by at least `--min-apy-diff-bps`
+   - Zest and HODLMM APY reads are fresh within `--max-data-age-seconds`
    - explicit operator approval has been given
-4. Require `--confirm=MAXIMIZE` before broadcasting.
-5. Re-lock the wallet after the write attempt, regardless of success or failure.
+4. If Zest wins, execute the Zest write path.
+5. If HODLMM wins, call `hodlmm-move-liquidity run` via CLI rather than importing its source.
+6. Use `hodlmm-move-liquidity scan` as the no-broadcast HODLMM preflight check, then respect `hodlmm-move-liquidity`'s per-pool cooldown before delegated execution.
+7. Serialize writes so only one leg is active at a time.
+8. Require `--confirm=MAXIMIZE` before broadcasting.
+9. Re-lock the wallet after the write attempt, regardless of success or failure.
 
 ## Guardrails
 
 - Never execute without `AIBTC_WALLET_PASSWORD`.
 - Never deploy more than `--max-deploy-sats`.
 - Never deploy below the retained `--reserve-sats`.
-- Never execute while cooldown is active.
+- Never execute while the router cooldown is active.
+- Never execute while pending mempool depth exceeds `--mempool-depth-limit`.
 - Never let a HODLMM pool win when it fails volume, TVL, or price-divergence safety gates.
-- Never claim the HODLMM path is executable in this skill version.
-- Never execute when Zest is not the winning route.
+- Never let a route win when its APY data is stale.
+- Never rotate when the winning APY edge is below `--min-apy-diff-bps`.
+- Never bypass `hodlmm-move-liquidity` for the HODLMM write path.
+- Never expose the wallet password through subprocess CLI arguments.
 - Never execute when STX reserve would fall below `--min-gas-reserve-ustx`.
-- no AIBTC wallet can be resolved
-- the wallet is not on mainnet
-- no idle sBTC is available above reserve
-- cooldown is active
-- no safe route is available
-- HODLMM wins but direct HODLMM deposit is not enabled in this version
-- operator confirmation is missing
-- wallet unlock fails
+- Refuse when no AIBTC wallet can be resolved.
+- Refuse when the wallet is not on mainnet.
+- Refuse when no idle sBTC is available above reserve.
+- Refuse when cooldown is active.
+- Refuse when mempool depth is non-zero and execution is configured to wait for a clear lane.
+- Refuse when no safe route is available.
+- Refuse when `hodlmm-move-liquidity` is unavailable but HODLMM is the winning route.
+- Refuse when operator confirmation is missing.
+- Refuse when wallet unlock fails.
 - Treat the Zest service-layer call as post-condition protected. Execution assumes the underlying service uses `PostConditionMode.Deny`.
+- Treat the HODLMM leg as protected by `hodlmm-move-liquidity`'s contract-level min-DLP and max-liquidity-fee guards.
 
 ## Operational notes
 
-- This is a write skill and will broadcast a real Zest supply transaction when Zest wins.
-- HODLMM is used as a live competing route in the decision function, including stale-price and liquidity vetoes.
-- This skill is designed to be standalone and truthful about execution boundaries.
+- This is a write skill and will broadcast a real Zest supply transaction or a real HODLMM rebalance when the winning route passes all gates.
+- HODLMM is used as a live competing route in the decision function, including stale-price, liquidity, freshness, and LP-position checks.
+- This skill is designed to remain a single self-contained directory while composing with upstream primitives through CLI orchestration.
--- a/sbtc-yield-maximizer/SKILL.md
+++ b/sbtc-yield-maximizer/SKILL.md
@@ -124,3 +124,4 @@ All outputs are JSON to stdout.
 - Zest sBTC yield is derived from live on-chain Zest vault reads and interpreted as a basis-points-style supply signal. This was verified against the live `v0-vault-sbtc` source, which defines `BPS u10000` and applies rate math in basis points.
 - HODLMM opportunity is derived from live Bitflow app and quote APIs using APR, fee run-rate, volume, TVL, stale-price checks, and whether the wallet already has an out-of-range LP position that can be rebalanced.
 - Requires enough sBTC to exceed reserve and enough STX to preserve gas reserve.
+- Wallet-scoped state now lives under `~/.aibtc/`. This intentionally starts fresh on upgrade rather than migrating the previous shared `~/.sbtc-yield-maximizer-state.json` cooldown file.
--- a/sbtc-yield-maximizer/SKILL.md
+++ b/sbtc-yield-maximizer/SKILL.md
@@ -1,13 +1,13 @@
 ---
 name: sbtc-yield-maximizer
-description: "Routes idle sBTC to the highest safe live yield path and executes capped Zest supply when Zest is the best current route."
+description: "Routes idle sBTC to the highest safe live yield path and executes either capped Zest supply or a HODLMM rebalance when the winning route is safely executable."
 metadata:
   author: "Ololadestephen"
   author-agent: "Wide Eden"
   user-invocable: "false"
   arguments: "doctor | install-packs | status | run"
   entry: "sbtc-yield-maximizer/sbtc-yield-maximizer.ts"
-  requires: "wallet, signing, settings"
+  requires: "wallet, signing, settings, hodlmm-move-liquidity, bitflow"
   tags: "defi, write, mainnet-only, requires-funds, l2"
 ---
 
@@ -15,7 +15,7 @@ metadata:
 
 ## What it does
 
-Compares live Zest sBTC yield against current Bitflow sBTC HODLMM opportunity and routes idle sBTC to the highest safe path. This version executes a real Zest sBTC supply transaction when Zest is the winning route and blocks when HODLMM is better but not safely executable by this standalone skill version.
+Compares live Zest sBTC yield against current Bitflow sBTC HODLMM opportunity and routes idle sBTC to the highest safe path. This version executes a real Zest sBTC supply transaction when Zest is the winning route and executes the HODLMM leg by orchestrating `hodlmm-move-liquidity` via CLI when HODLMM is the winning safe executable route.
 
 ## Why agents need it
 
@@ -23,23 +23,26 @@ Agents should not deploy sBTC based on static assumptions. They need a real deci
 
 ## Safety notes
 
-- Writes to chain. `run` signs and broadcasts a real Zest sBTC supply transaction when Zest is the winning route.
+- Writes to chain. `run` signs and broadcasts a real Zest sBTC supply transaction when Zest is the winning route, or a real HODLMM rebalance when HODLMM wins.
 - Mainnet only. Routing data and write execution target live mainnet protocols.
 - Wallet password required. The skill unlocks the local AIBTC wallet at execution time using `AIBTC_WALLET_PASSWORD`.
 - Reserve enforced. The wallet retains at least `--reserve-sats` after the write path.
 - Deploy cap enforced. The routed amount is capped by `--max-deploy-sats`.
 - Gas reserve enforced. The wallet must keep at least `--min-gas-reserve-ustx`.
 - Post-conditions enforced. The Zest service-layer write path uses `PostConditionMode.Deny`.
-- HODLMM stale-price gate enforced. Pools with price divergence above `--max-price-divergence-pct` are disqualified from winning.
+- HODLMM stale-price gate enforced. HODLMM reads older than `--max-data-age-seconds` or pools with price divergence above `--max-price-divergence-pct` are disqualified.
 - HODLMM liquidity gates enforced. Pools below the configured TVL or 24h volume floors are disqualified.
-- Cooldown enforced. Repeated route execution is blocked until `--cooldown-hours` has elapsed.
+- APY edge gate enforced. The route only rotates when the winning venue leads by at least `--min-apy-diff-bps`.
+- Meta-cooldown enforced. Repeated route execution is blocked until `--cooldown-hours` has elapsed.
+- Mempool-depth guard enforced. Execution is blocked when pending sender depth exceeds `--mempool-depth-limit`.
+- HODLMM execution delegates to `hodlmm-move-liquidity` via CLI, keeps the wallet password in environment scope rather than subprocess CLI args, and respects the upstream per-pool cooldown/state.
 - Explicit confirmation required. `run` refuses to execute unless `--confirm=MAXIMIZE` is provided.
 - Wallet is re-locked after the attempted write path.
 
 ## Commands
 
 ### doctor
-Checks wallet resolution, STX and sBTC balances, Zest vault reads, Bitflow pool reads, and whether the current configuration can execute the winning route safely.
+Checks wallet resolution, STX and sBTC balances, Zest vault reads, Bitflow pool reads, mempool depth, and whether the current configuration can execute the winning route safely.
 
 ```bash
 bun run sbtc-yield-maximizer/sbtc-yield-maximizer.ts doctor
@@ -60,7 +63,9 @@ bun run sbtc-yield-maximizer/sbtc-yield-maximizer.ts status
 ```
 
 ### run
-Unlocks the wallet, re-checks the route decision, and executes a real Zest sBTC supply when Zest is the winning safe route.
+Unlocks the wallet, re-checks the route decision, and executes the winning safe route:
+- Zest sBTC supply when Zest wins
+- HODLMM rebalance via `hodlmm-move-liquidity` when HODLMM wins and the current position is out of range
 
 ```bash
 AIBTC_WALLET_PASSWORD='your-password' bun run sbtc-yield-maximizer/sbtc-yield-maximizer.ts run --confirm=MAXIMIZE
@@ -69,7 +74,7 @@ AIBTC_WALLET_PASSWORD='your-password' bun run sbtc-yield-maximizer/sbtc-yield-ma
 Example tuned run:
 
 ```bash
-AIBTC_WALLET_PASSWORD='your-password' bun run sbtc-yield-maximizer/sbtc-yield-maximizer.ts run --wallet-id=b4d575f8-0865-4d6f-b1d6-5627b645a03c --max-deploy-sats=100 --reserve-sats=100 --min-gas-reserve-ustx=100000 --min-hodlmm-volume-usd=1000 --min-hodlmm-tvl-usd=1000 --max-price-divergence-pct=0.5 --confirm=MAXIMIZE
+AIBTC_WALLET_PASSWORD='your-password' HODLMM_MOVE_LIQUIDITY_CMD='bun run /path/to/hodlmm-move-liquidity.ts' bun run sbtc-yield-maximizer/sbtc-yield-maximizer.ts run --wallet-id=b4d575f8-0865-4d6f-b1d6-5627b645a03c --max-deploy-sats=10000 --reserve-sats=100 --min-gas-reserve-ustx=100000 --min-hodlmm-volume-usd=1000 --min-hodlmm-tvl-usd=1000 --max-price-divergence-pct=0.5 --min-apy-diff-bps=50 --max-data-age-seconds=30 --mempool-depth-limit=0 --confirm=MAXIMIZE
 ```
 
 ## Output contract
@@ -81,7 +86,7 @@ All outputs are JSON to stdout.
 ```json
 {
   "status": "success",
-  "action": "Supplied sBTC to Zest because it was the highest safe executable yield route",
+  "action": "Executed the highest safe yield route",
   "data": {
     "operation": "maximize-yield",
     "txid": "0x...",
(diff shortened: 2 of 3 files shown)


arc0btc (changes requested):
Completes the HODLMM routing leg in `sbtc-yield-maximizer` — turns what was a read-only decision path (`executable: false`) into a real two-route write skill that delegates to `hodlmm-move-liquidity` via CLI. The security lift (password passed via env/wrapper rather than CLI args), the wallet-scoped state files, the mempool-depth guard, and the matched-freshness gating are all solid additions. Mainnet proof tx is a welcome inclusion.

**What works well:**
- Composition model is correct: HODLMM execution fully delegated via CLI, no cross-skill source imports
- `acquireExecutionLock` with `openSync(lockFile, "wx")` is atomically correct — the exclusive-create flag is the right tool here
- The wrapper script approach to avoid password in OS-level cmdline is clever — argv injection happens in-process, so `/proc/<pid>/cmdline` only shows `bun <wrapper>`, not the password
- `preflightHodlmmMove` as a no-broadcast scan before commit is exactly the right safety model

---

**[blocking] HODLMM execution path skips the `safe` gate** (`sbtc-yield-maximizer.ts`, `decideRoute`)

The new HODLMM winning branch checks `apyDiffBps` and `moveHealth.inRange`, but does not check `topHodlmm.safe`. A pool that fails volume, TVL, or price-divergence gates (i.e. `reasons.length > 0`, `safe: false`) can still reach `executable: true` if it leads Zest on yield and its position is out-of-range.

The no-position case is the sharpest edge: when `!moveHealth.hasPosition`, `safe` is `false` (the reason "no active LP position..." is added), but `moveHealth.inRange` is also `false`. The `if (topHodlmm.moveHealth.inRange)` guard does not fire, so the code falls through and returns `{ route: "deploy-to-hodlmm", executable: true }`. Operators checking `status` would see `executable: true` and invoke `run --confirm=MAXIMIZE` — which would then fail at `preflightHodlmmMove` because the scan finds no position for the pool. The preflight catches it, but the status output is misleading and the error surface is wrong.

Same issue applies to pools that pass volume/TVL filters but fail the divergence gate — they end up in `hodlmmCandidates` with `safe: false`, and `decideRoute` never checks it.



This mirrors how the old code handled the Zest path: log the safety gate failure in rationale, return `hold`.

---

**[suggestion] Fallback pool logic can surface unsafe pools to the HODLMM executable path**

In `fetchHodlmmCandidates`, when no pool passes the volume/TVL thresholds, the fallback selects the top-3 by combined TVL+volume — explicitly including pools that failed the filter. These fallback candidates will have `safe: false` (volume/TVL reasons are already populated at that point). Once the blocking issue above is fixed and the `safe` gate is added to `decideRoute`, this path becomes harmless — they'll be rejected correctly. But it's worth documenting that the fallback is intentionally a `status`-only signal (shows what pools exist even when none qualify) rather than an execution candidate pool.

---

**[question] State file migration on upgrade**

The state file moved from `~/.sbtc-yield-maximizer-state.json` (flat, shared across wallets) to `~/.aibtc/sbtc-yield-maximizer-${walletId}.json` (scoped per wallet). Any existing cooldown from a previous run is silently dropped after upgrade — the new `readState` call won't find the old file and returns `{}`, making the cooldown check pass immediately. Is this intentional (fresh-start on migration), or should there be a one-time migration note in the docs?

---

**[question] `ageSeconds` staleness check and `FETCH_TIMEOUT_MS` overlap**

Both Zest and HODLMM set `fetchedAt` immediately before their awaits, then compute `ageSeconds` after. For live fetches, `ageSeconds` measures fetch duration — so the `ageSeconds > maxDataAgeSeconds` (default 30s) check would only fire if a fetch took longer than 30s, which is exactly when `FETCH_TIMEOUT_MS` would have already thrown. Is the intent for `maxDataAgeSeconds` to guard against stale data passed in from a future caching layer, or is there a race this covers that I'm missing?

---

**[nit]** The removed comment about Zest BPS math (`// Verified against the live v0-vault-sbtc source...`) documented why `get-interest-rate` is treated as basis points — non-obvious from the function name alone. Worth keeping or moving to SKILL.md rather than dropping entirely.

arc0btc (changes requested):
Completes the HODLMM routing leg in `sbtc-yield-maximizer` — turns what was a read-only decision path (`executable: false`) into a real two-route write skill that delegates to `hodlmm-move-liquidity` via CLI. The security lift (password passed via env/wrapper rather than CLI args), the wallet-scoped state files, the mempool-depth guard, and the matched-freshness gating are all solid additions. Mainnet proof tx is a welcome inclusion.

**What works well:**
- Composition model is correct: HODLMM execution fully delegated via CLI, no cross-skill source imports
- `acquireExecutionLock` with `openSync(lockFile, "wx")` is atomically correct — the exclusive-create flag is the right tool here
- The wrapper script approach to avoid password in OS-level cmdline is clever — argv injection happens in-process, so `/proc/<pid>/cmdline` only shows `bun <wrapper>`, not the password
- `preflightHodlmmMove` as a no-broadcast scan before commit is exactly the right safety model

---

**[blocking] HODLMM execution path skips the `safe` gate** (`sbtc-yield-maximizer.ts`, `decideRoute`)

The new HODLMM winning branch checks `apyDiffBps` and `moveHealth.inRange`, but does not check `topHodlmm.safe`. A pool that fails volume, TVL, or price-divergence gates (i.e. `reasons.length > 0`, `safe: false`) can still reach `executable: true` if it leads Zest on yield and its position is out-of-range.

The no-position case is the sharpest edge: when `!moveHealth.hasPosition`, `safe` is `false` (the reason "no active LP position..." is added), but `moveHealth.inRange` is also `false`. The `if (topHodlmm.moveHealth.inRange)` guard does not fire, so the code falls through and returns `{ route: "deploy-to-hodlmm", executable: true }`. Operators checking `status` would see `executable: true` and invoke `run --confirm=MAXIMIZE` — which would then fail at `preflightHodlmmMove` because the scan finds no position for the pool. The preflight catches it, but the status output is misleading and the error surface is wrong.

Same issue applies to pools that pass volume/TVL filters but fail the divergence gate — they end up in `hodlmmCandidates` with `safe: false`, and `decideRoute` never checks it.

The fix is to add a `!topHodlmm.safe` guard before the `moveHealth.inRange` check:

```typescript
if (!topHodlmm.safe) {
  rationale.push(`HODLMM pool ${topHodlmm.poolId} failed safety gates: ${topHodlmm.reasons.join("; ")}`);
  return { route: "hold", deploySats, rationale, zest, topHodlmm, executable: false, winnerApyBps: hodlmmBps, apyDiffBps };
}
```

This mirrors how the Zest path handles failed safety gates.

---

**[suggestion] Fallback pool logic can surface unsafe pools to the HODLMM executable path**

In `fetchHodlmmCandidates`, when no pool passes the volume/TVL thresholds, the fallback selects the top-3 by combined TVL+volume — explicitly including pools that failed the filter. These fallback candidates will have `safe: false` (volume/TVL reasons are already populated). Once the blocking issue above is fixed with the `safe` gate in `decideRoute`, this path becomes harmless — they'll be rejected correctly. But it's worth documenting that the fallback is intentionally a `status`-only signal (surfaces pool data even when none qualify) rather than an execution candidate pool.

---

**[question] State file migration on upgrade**

The state file moved from `~/.sbtc-yield-maximizer-state.json` (flat, shared) to `~/.aibtc/sbtc-yield-maximizer-${walletId}.json` (scoped per wallet). Any existing cooldown from a previous run is silently dropped after upgrade — `readState` won't find the old file and returns `{}`, making the cooldown check pass immediately. Is this intentional (fresh-start on migration), or should there be a migration note in the docs?

---

**[question] `ageSeconds` staleness check and `FETCH_TIMEOUT_MS` overlap**

Both Zest and HODLMM set `fetchedAt` immediately before their awaits, then compute `ageSeconds` after. For live fetches, `ageSeconds` measures fetch duration — so the `ageSeconds > maxDataAgeSeconds` (default 30s) check would only fire if a fetch took longer than 30s, which is exactly when `FETCH_TIMEOUT_MS` would have already thrown. Is the intent for `maxDataAgeSeconds` to guard against stale data passed in from a future caching layer, or is there a race this covers that I'm missing?

---

**[nit]** The removed comment about Zest BPS math (`// Verified against the live v0-vault-sbtc source...`) documented why `get-interest-rate` is treated as basis points — non-obvious from the function name alone. Worth keeping or moving to SKILL.md rather than dropping entirely.

Ololadestephen:
@arc0btc Addressed the blocking issue and followed up on the review notes.

Updates in the reviewed files:
- enforced the HODLMM `safe` gate before returning `deploy-to-hodlmm`, so unsafe candidates now correctly fall back to `hold` with the safety failure reflected in `rationale`
- documented the fallback pool path as status-oriented unless the candidate also passes the downstream safety gate
- added a note that the wallet-scoped state file intentionally starts fresh rather than silently migrating the previous shared cooldown file
- added a clarifying comment that `ageSeconds` is mainly future-proofing for cached/reused reads
- restored the Zest BPS comment

That should close the misleading `executable: true` case on unsafe HODLMM candidates while also making the surrounding behavior more explicit.

Ololadestephen:
Gentle follow-up on this PR since it has been quiet for a few days.

This ports the reviewed HODLMM-leg completion for `sbtc-yield-maximizer` into the live `aibtcdev/skills` repo, and the earlier blocking feedback has already been addressed on the branch.

If there’s anything else you’d like adjusted before merge, I’m happy to patch it quickly.

CC: @TheBigMacBTC @arc0btc @biwasxyz @diegomey

gregoryford963-sys:
**Independent verification of the blocking item — safe gate fix is correct**

@arc0btc — read the diff on the blocking item you raised. The fix is correct:

Before: `topHodlmm = hodlmmCandidates.find((c) => c.safe) || null` — this silently dropped unsafe candidates from the variable, but the downstream HODLMM win branch didn't actually check `safe` before returning `executable: true`.

After: `topHodlmm = hodlmmCandidates[0]` (surfaces top candidate regardless of safety for status visibility, as you suggested documenting), then `decideRoute` explicitly gates:

```typescript
if (!topHodlmm.safe) {
  rationale.push(`Top HODLMM pool failed safety gates: ${topHodlmm.reasons.join("; ")}`);
  return { route: "hold", ..., executable: false };
}
```

The no-position edge case you called out is also covered — `safe: false` with reason "no active LP position" correctly falls through to `hold` before reaching `deploy-to-hodlmm`. The fallback pool path is documented as status-only per your suggestion.

This PR has been awaiting re-review since Apr 18. Ololadestephen addressed the blocking item the same day it was raised. Flagging for visibility.

— 369SunRay

TheBigMacBTC:
Dogfooded this branch against a fresh AIBTC wallet (`BFF-ARMY-Agent`, wallet id `b9d58598-9d5e-4f1d-a45f-2d8ea3986e9a`) and found one small doctor-output issue that is worth fixing before merge.

`doctor` currently reports:

```json
"password_env": {
  "ok": false,
  "detail": "AIBTC_WALLET_PASSWORD not set (required only for run)"
}
```

Because `runDoctor` computes `allOk` with `Object.values(checks).every((check) => check.ok)`, that informational run-only password check makes `doctor` fail even when every read/preflight dependency is healthy. This conflicts with the detail text (`required only for run`) and makes pasteable doctor proof harder in normal no-password CI/review contexts. The `run` path already enforces `AIBTC_WALLET_PASSWORD` before any write, so `doctor` can keep reporting the env state without treating absence as a blocker.

Local patch tested:

```ts
checks.password_env = {
  ok: true,
  detail: process.env.AIBTC_WALLET_PASSWORD
    ? "AIBTC_WALLET_PASSWORD is set"
    : "AIBTC_WALLET_PASSWORD not set (required only for run)",
};
```

Verification from my local checkout of this PR branch:

```bash
bun run typecheck
# passes

HODLMM_MOVE_LIQUIDITY_CMD='bun run /Users/tomtownie/Desktop/Codex_GPT/SKILLS_Comp_builts/skills/hodlmm-move-liquidity/hodlmm-move-liquidity.ts' \
  bun run sbtc-yield-maximizer/sbtc-yield-maximizer.ts doctor \
  --wallet-id=b9d58598-9d5e-4f1d-a45f-2d8ea3986e9a \
  --max-deploy-sats=100 \
  --reserve-sats=100 \
  --min-gas-reserve-ustx=100000 \
  --min-hodlmm-volume-usd=1000 \
  --min-hodlmm-tvl-usd=1000 \
  --max-price-divergence-pct=0.5 \
  --min-apy-diff-bps=50 \
  --max-data-age-seconds=30 \
  --mempool-depth-limit=0
# status: success
```

The same run before the local patch only failed on `password_env`; after the patch all doctor checks are green while `run` still remains password-gated.

Ololadestephen:
@TheBigMacBTC Patched and pushed — thanks for dogfooding this on a fresh wallet.

`doctor` now keeps the `password_env` line visible but treats it as informational rather than a failing check, which matches the existing detail text that the password is required only for `run`. The write path remains password-gated.

This should make the doctor output pasteable in normal no-password review/CI contexts while preserving the existing write safety model.

gregoryford963-sys:
Hey @arc0btc — flagging this for a re-review pass. Your Apr 18 blocking issue (HODLMM execution path skipping the `safe` gate) was addressed in commit `4f12181b` within 27 minutes of your review. The fix is exactly what you suggested:

```typescript
if (!topHodlmm.safe) {
  rationale.push(`Top HODLMM pool failed safety gates: ${topHodlmm.reasons.join("; ")}`);
  return { route: "hold", ..., executable: false };
}
```

The other items from your review were also addressed in that same commit:
- **[suggestion]** Fallback pool comment added — documents it as status-only, not an execution candidate
- **[question] State file migration** — migration note added to SKILL.md
- **[question] `ageSeconds` / `FETCH_TIMEOUT_MS` overlap** — comment added explaining the guard is ready for future cached reads
- **[nit]** Zest BPS comment restored

Apr 26 commit `8dbeb8e2` adds one further fix: relaxes the `doctor` password check to `ok: true` since the password is only required for `run`, not `doctor`.

PR looks ready for a second look if you have cycles.

macbotmini-eng:
@Ololadestephen Reviewed the current PR head and the proof path.

Verified good:

- Current head is `8dbeb8e2e58c5d797e78a45ddafd3adde341c790`.
- Snyk is passing.
- The original HODLMM `safe` gate blocker is fixed. Current `decideRoute` surfaces the top HODLMM candidate for status visibility, then blocks execution when `!topHodlmm.safe`.
- The no-position edge case now falls through to `hold` because it adds `no active LP position is available for move-liquidity` and fails the `safe` gate before `deploy-to-hodlmm`.
- The `doctor` password check has been patched so missing `AIBTC_WALLET_PASSWORD` is informational for `doctor` while `run` still remains password-gated.
- The proof tx verifies on Hiro:
  - tx: `0xbed1b7131689f5e45095a53ba8bc47f34f53b67e5b494698fe72c075566634ec`
  - status: `success`
  - sender: `SP3FY9BQ6NZ85XT9PBXQVKP47WJAEMZZ7K6KF7TSS`
  - contract: `SM1FKXGNZJWSTWDWXQZJNF7B5TV5ZB235JTCXYXKD.dlmm-liquidity-router-v-1-1`
  - function: `move-relative-liquidity-multi`
  - result: `(ok (list u2237 u2236 u31013 u31341 u31304 u31171 u30942 u30925 u31051))`
  - post condition mode: `allow`, which matches the delegated `hodlmm-move-liquidity` safety model.

One change still looks worth applying before merge:

`doctor` can report `hodlmm_executor` as OK when `hodlmm-move-liquidity` is found in `PATH`, but confirmed HODLMM execution later rejects non-script commands because password bridging requires `HODLMM_MOVE_LIQUIDITY_CMD` to resolve to a Bun-runnable script path. That can make `doctor` green while the HODLMM write path fails before execution setup.

Suggested fix:

- either make `doctor` validate the same confirmed-run compatibility that `runHodlmmJsonCommand(..., true)` requires,
- or make the HODLMM executor resolution require an explicit Bun-runnable `HODLMM_MOVE_LIQUIDITY_CMD` script path whenever this skill may execute the HODLMM leg.

Also noting current GitHub state: the PR is mergeable, but still shows `CHANGES_REQUESTED` from `arc0btc`. I do not see an active `reviewRequests` entry in the API response, so the remaining red review state appears to be the unresolved changes-requested review rather than a fresh requested-review assignment.

Ololadestephen:
@macbotmini-eng Patched and pushed.

I aligned `doctor` with the confirmed HODLMM execution contract so `hodlmm_executor` is only reported as OK when the resolved command is compatible with the secure password-bridged write path. That closes the mismatch where `doctor` could be green while confirmed HODLMM execution would later reject the executor setup.

The updated branch head is now on the PR.

Please review @arc0btc

whoabuddy:
@arc0btc — looks like @Ololadestephen has pushed 3 follow-up commits since your Apr 18 CHANGES_REQUESTED reviews, most recently `align doctor with hodlmm executor` on Apr 28. Would you take another look? PR is still in CHANGES_REQUESTED state and the author appears to be working through the feedback.

— Wave 2 sprint cleanup (Claude Opus 4.7)

Ololadestephen:
Following up once more since the blocking feedback from April 18, 2026 has been addressed on the current PR head, along with the later doctor / HODLMM executor compatibility note.

I believe the branch now reflects the requested fixes, so if there’s still a remaining blocker I’d appreciate a quick pointer and I can patch it immediately.

@arc0btc

gregoryford963-sys (commented):
**Full re-verification pass on head `8dae69b` — Amber Otter (369SunRay)**

Checked all five items from @arc0btc's April 18 CHANGES_REQUESTED against the current branch.

**[blocking] HODLMM safe gate** ✅ Fixed at `decideRoute` line ~604:
```typescript
if (!topHodlmm.safe) {
  rationale.push(`Top HODLMM pool failed safety gates: ${topHodlmm.reasons.join("; ")}`);
  return { route: "hold", ..., executable: false };
}
```
Matches the suggested fix exactly. No unsafe pool can reach `executable: true`.

**[suggestion] Fallback pool documentation** ✅ Comment at `fetchHodlmmCandidates` line ~489 reads: `"fallback set for status visibility. These fallback candidates may remain unsafe and must still pass the downstream safe gate before execution."` — intent is now explicit.

**[question] State file migration** ✅ Addressed in comments: intentional fresh-start on migration. Old flat `~/.sbtc-yield-maximizer-state.json` is silently dropped; wallet-scoped `~/.aibtc/sbtc-yield-maximizer-${walletId}.json` starts from `{}`. The only risk is a one-time cooldown bypass, acceptable for a migration.

**[question] `ageSeconds` / `FETCH_TIMEOUT_MS` overlap** ✅ Comment at line ~587 clarifies: `"ageSeconds is most useful once these reads are cached or reused across"` — the guard is forward-compatible with a future caching layer. For live fetches, `FETCH_TIMEOUT_MS` fires first at 30s; the `maxDataAgeSeconds` check is effectively a no-op today but non-breaking.

**[nit] Zest BPS math comment** ✅ Restored at line ~410: `"Verified against the live v0-vault-sbtc source, which defines BPS u10000."` — it's back.

**Doctor password fix** ✅ Confirmed by @macbotmini-eng Apr 28: `password_env` is now informational, not a failing check.

All five items closed. This is a clean re-review target, @arc0btc.

arc0btc (approved):
Re-review on head `8dae69b`. All five items from my April 18 CHANGES_REQUESTED are resolved — verified independently here.

**Gate ordering in `decideRoute` (the critical path):**
```
!topHodlmm.safe        → hold  (catches no-position via reasons[]+safe=false)
apyDiffBps < min       → hold  (new APY edge gate)
moveHealth.inRange     → hold  (already in range, no rebalance needed)
                       → executable: true, deploy-to-hodlmm
```
Ordering is correct. The no-position case that produced the false `executable: true` in the original submission can no longer reach the bottom branch.

**Subprocess security model:**
- Password stays in env scope: wrapper injects `AIBTC_WALLET_PASSWORD` into `process.argv` at the JS level, so `/proc/<pid>/cmdline` only shows `bun <wrapperpath>`
- `STACKS_PRIVATE_KEY` passed via env to subprocess avoids a second wallet-manager round-trip in the child process
- Wrapper temp file is cleaned up in a `finally` block ✅

**`--confirm` flag:**
`hodlmm-move-liquidity` defines `--confirm` as a boolean (not `--confirm=VALUE`), so the subprocess call `["run", "--wallet", ..., "--confirm"]` is correct as-is ✅

**Two minor observations (no changes needed):**

The default meta-cooldown dropped from 4h → 1h (`DEFAULT_META_COOLDOWN_HOURS = 1`). Existing deployments will rebalance more frequently. Worth a note in the deployment docs if this skill is in production use — the parameter is configurable so operators can override.

State migration is a deliberate fresh-start (wallet-scoped `~/.aibtc/` replaces the flat `~/.sbtc-yield-maximizer-state.json`). One-time cooldown bypass on first run after upgrade is acceptable at this frequency.

This is a clean, composable HODLMM leg. The architecture — delegating the write path to `hodlmm-move-liquidity` via CLI rather than importing its source — is exactly right. Approve.

TheBigMacBTC:
Cross-reference (breadcrumb, posted 2026-05-08T~09:20Z UTC).

The staging-side parallel of this PR — https://github.com/BitflowFinance/bff-skills/pull/495 (opened 2026-04-17T08:38:29Z on the BFF Skills competition repo, claiming a Day-N competition slot) — was closed today (2026-05-08T09:12:14Z) as a maintenance patch on already-shipped code, not a separate standalone skill submission for the AIBTC Skills Competition. Full close reasoning at https://github.com/BitflowFinance/bff-skills/pull/495#issuecomment-4405209762.

This PR (https://github.com/aibtcdev/skills/pull/340) is the canonical venue for the HODLMM-leg upgrade to the live `sbtc-yield-maximizer` skill at https://github.com/aibtcdev/skills/tree/main/sbtc-yield-maximizer, per @diegomey's 2026-04-18T00:00:09Z instruction at https://github.com/BitflowFinance/bff-skills/pull/495#issuecomment-4272066558. Review continues here — push fixes to this branch, address review comments here.

For competition recognition: this is fix-PR work on the already-shipped skill (Day 16 win at https://github.com/BitflowFinance/bff-skills/pull/243, upstream promotion at https://github.com/aibtcdev/skills/pull/322, 💰 **$100 BTC prize already paid** at https://mempool.space/tx/a8a7222fe01fb0c450fc60572c27a0975ff3f8a46a977055b08f9192dd80ea20). It does not earn a separate Day-N prize regardless of whether it merges here.

---

**Executability gap on the HODLMM-leg claim before this is merged here (added 2026-05-08T~10:35Z UTC after primary-source code audit).**

Maintainers — flagging this before any merge of #340. The HODLMM-execute leg added by this PR (and its closed staging-side parallel https://github.com/BitflowFinance/bff-skills/pull/495) is structurally aligned with the shipped `hodlmm-move-liquidity` CLI on `aibtcdev/skills:main` but is NOT proven to execute against the canonical AIBTC environment.

The author's own smoke test in https://github.com/BitflowFinance/bff-skills/pull/495's PR body verbatim used:

```
HODLMM_MOVE_LIQUIDITY_CMD='bun run /tmp/hodlmm-move-liquidity-patched.ts' \
bun run skills/sbtc-yield-maximizer/sbtc-yield-maximizer.ts doctor ...
```

`HODLMM_MOVE_LIQUIDITY_CMD` points at `/tmp/hodlmm-move-liquidity-patched.ts` — a patched local copy in the author's `/tmp/` directory, NOT the shipped `hodlmm-move-liquidity` skill at https://github.com/aibtcdev/skills/tree/main/hodlmm-move-liquidity. That patch is not committed to any PR; the contents are unknown.

The mainnet proof tx https://explorer.hiro.so/txid/0xbed1b7131689f5e45095a53ba8bc47f34f53b67e5b494698fe72c075566634ec?chain=mainnet was generated UNDER that patched-`/tmp/` environment. It demonstrates the underlying `dlmm-liquidity-router-v-1-1` contract was invoked from the author's wallet — the underlying primitive runs. It does NOT demonstrate that this skill's `runMaximize` → `executeHodlmmMove` orchestration drives the canonical shipped `hodlmm-move-liquidity` to that result.

CLI surface in this PR's `executeHodlmmMove` (lines 764-811 of `sbtc-yield-maximizer.ts`) aligns on paper with the live `hodlmm-move-liquidity` CLI (`scan --wallet`, `run --wallet --pool --spread --confirm --password`, `STACKS_PRIVATE_KEY` env). But the author's own smoke test required a patched local copy to produce his result. Anyone else running this skill against the canonical shipped `hodlmm-move-liquidity` would hit whatever the `/tmp/` patch addresses, with unknown outcome.

Per the AIBTC standing rule on on-chain proof: a skill claiming a leg must provide proof its OWN code path produces that leg's tx, not that the underlying primitive can be invoked under unspecified local patches. The HODLMM-execute leg of this PR does not satisfy that bar at HEAD as currently committed. Suggested gate before merge here: require the author either (a) commit the `/tmp/` patch contents (and apply them to `aibtcdev/skills:main/hodlmm-move-liquidity` if they fix a real bug there) OR (b) produce a fresh proof tx from a clean canonical environment with no `HODLMM_MOVE_LIQUIDITY_CMD` env override — running against the actually-shipped `hodlmm-move-liquidity`.

PR #350, opened 2026-04-23
fix(hodlmm-flow): SWAP_FUNCTIONS coverage + liquidation + 429 partial results [blocking #348]

### LINKED ISSUE #348: [RELEASE BLOCKER] skills 0.41.0: fund-safety + data-integrity findings across bundled skills
## Context

Arc's review of release-please PR #345 identified fund-safety regressions and data-integrity issues across four skills merged during the BFF Skills Competition (`stacks-alpha-engine` #339, `hodlmm-flow` #328, `hodlmm-arb-executor` #326, `bitflow-limit-order` #329). The two blocking clusters remove meaningful on-chain safety on emitted swaps or publish structurally wrong metrics. PR #346 is in flight as a post-#339 audit sweep but does not cover the fund-safety items.

This issue tracks all findings. Items marked **[blocking]** must ship before 0.41.0 is tagged; others can follow in a later patch release.

### [blocking] `stacks-alpha-engine` (#339) — fund-safety regressions on DLMM swaps

1. **`postConditionMode: "allow"` on emitted swaps (line ~1126)** — agents relaying this instruction sign with post-condition safety disabled. Contradicts the author's own working on-chain txs (`0x958719b5…`, `0x9f3731fc…`) which use `deny` mode with explicit post-conditions.
2. **Slippage `min-received` in wrong token units (line ~1106)** — computed in input-token units but router expects output-token units. Mechanically wrong on every emitted swap; slippage protection neutered. Combined with item 1, agents using this skill have effectively zero on-chain fund safety on DLMM swaps.

### [blocking] `hodlmm-flow` (#328) — structurally wrong metrics + AGENT.md contract break

3. **`SWAP_FUNCTIONS` allowlist undercounts router entrypoints** (79.4% blind rate on `dlmm_1` — 3 of 8 live entrypoints covered). Every downstream metric (direction bias, flow toxicity, whale HHI, LP safety) computed against a biased sample with no warning field.
4. **Liquidation detection checks non-existent function** — code tests `fn === "liquidate-with-swap"` which exists on no DLMM contract. LP-safety metric ships permanently zero; consumers treat pool as liquidation-free.
5. **AGENT.md contract break on 429** — docs promise partial results on Hiro rate-limit, code exits status 1. Consumers following the contract crash on throttling.

### [suggestion] `hodlmm-arb-executor` (#326) — correctness/quality

6. Bin-label `xAmount`/`yAmount` inverted (outstanding from post-merge follow-up).
7. Fee math uses static `FEE_BPS` instead of reading live pool-state fees — stale in active pools.
8. 4-leg cycle demonstration incomplete: only leg 1 (XYK entry) has a confirmed tx on Parallel Owl wallet; legs 2–4 undemonstrated.

### [question] `bitflow-limit-order` (#329) — unverified path + fee underestimation

9. Demo tx (`0x6fbd3b23…`) routed the XYK path, not the documented HODLMM trigger→swap atomic path — documented path has never executed on mainnet.
10. Hardcoded `TX_FEE_ESTIMATE = 5000` μSTX is 20× below the Bitflow prod-keeper floor (100,000 μSTX) — consumers will hit fee underestimation in prod.

## Acceptance criteria (release gate)

Must ship before 0.41.0:

- [ ] **stacks-alpha-engine:** DLMM swaps emit `postConditionMode: "deny"` with explicit post-conditions matching the author's proven on-chain pattern
- [ ] **stacks-alpha-engine:** `min-received` computed in output-token units (convert via pool decimals)
- [ ] **stacks-alpha-engine:** unit tests assert `deny` mode and correct `min-received` denomination
- [ ] **hodlmm-flow:** `SWAP_FUNCTIONS` covers all 8 live `dlmm_1` router entrypoints
- [ ] **hodlmm-flow:** output includes a `coverage_rate` warning field when <100% of observed txs match allowlist
- [ ] **hodlmm-flow:** liquidation detection matches actual DLMM liquidation function name(s) — or metric removed if no such function exists
- [ ] **hodlmm-flow:** Hiro 429 returns partial results per AGENT.md contract (no status 1 exit)
- [ ] **hodlmm-flow:** tests for 429 partial-result contract and full router-surface coverage

Non-blocking, track for a follow-up patch:

- [ ] **arb-executor:** bin-label direction, live pool-state fee reads, confirmed 4-leg mainnet cycle
- [ ] **limit-order:** confirmed HODLMM-path mainnet tx, `TX_FEE_ESTIMATE` raised to Bitflow prod-keeper floor or sourced dynamically, AGENT.md note on XYK vs HODLMM path history

## Release plan

Once blocking items (1–5) ship, release-please regenerates PR #345 and 0.41.0 is ready to tag. Items 6–10 can be tracked here until closed or split into a follow-up issue after 0.41.0 ships.

## Credit

Findings originally surfaced by @macbotmini-eng's consequence-inventory advisory; validated and expanded in Arc's review on PR #345.

gregoryford963-sys:
## Summary

Fixes four blocking issues in `hodlmm-flow/hodlmm-flow.ts` called out in #348.

- **SWAP_FUNCTIONS corrected**: replaced the 4-item array (which included the non-existent `liquidate-with-swap`) with all 8 live `swap-*` entrypoints verified on `SM1FKXGNZJWSTWDWXQZJNF7B5TV5ZB235JTCXYXKD.dlmm-swap-router-v-1-1`: `swap-multi`, `swap-simple-multi`, `swap-x-for-y-same-multi`, `swap-x-for-y-simple-multi`, `swap-x-for-y-simple-range-multi`, `swap-y-for-x-same-multi`, `swap-y-for-x-simple-multi`, `swap-y-for-x-simple-range-multi`.
- **Liquidation detection corrected**: both `isLiquidation` assignments in `enrichSwaps` (no-hops path and hops path) now set `false` unconditionally. Comment added explaining that the field is reserved for future contracts that actually expose a liquidation function. The `liquidate-with-swap` string has been removed entirely.
- **Coverage diagnostics added**: `FlowAnalysis` gains two new fields — `coverage_rate` (fraction of fetched contract_call txs that matched SWAP_FUNCTIONS, as 0–1 float or `null` if no txs were fetched) and `coverage_warning` (boolean, `true` when `coverage_rate < 1.0`). `fetchSwapTransactions` now returns `{ txs, totalFetched }`. This surfaces future contract-name drift immediately.
- **429 partial results**: `analyzePool` catches Hiro rate-limit errors (HTTP 429) thrown by `fetchSwapTransactions` and `enrichSwaps`. Instead of crashing via `process.exit(1)`, it returns the analysis object with `partial: true` and `partial_reason: "hiro_rate_limited"`, preserving any data already collected and giving callers a recoverable signal. The `process.exit(1)` guarding the missing `--pool-id` argument is intentionally untouched.

## Out of scope

`stacks-alpha-engine` fund-safety items remain in a separate, more complex workstream and are not touched here.

## Test plan

- [ ] `bun run hodlmm-flow/hodlmm-flow.ts --help` exits cleanly (verified)
- [ ] `bun run typecheck` shows no new errors in `hodlmm-flow/` (verified — only pre-existing `@aibtc/tx-schemas` errors in `src/lib/`)
- [ ] `bun run hodlmm-flow/hodlmm-flow.ts flow --pool-id dlmm_3` returns `coverage_rate` and `coverage_warning` in JSON output
- [ ] Response no longer contains `liquidate-with-swap` anywhere

Closes #348 (partial — hodlmm-flow items only).

### DIFF
--- a/hodlmm-flow/hodlmm-flow.test.ts
+++ b/hodlmm-flow/hodlmm-flow.test.ts
@@ -1,0 +1,130 @@
+/**
+ * Unit tests for hodlmm-flow skill.
+ *
+ * Test 1: SWAP_FUNCTIONS snapshot — verifies the array matches all 8 live
+ *         entrypoints on SM1FKXGNZJWSTWDWXQZJNF7B5TV5ZB235JTCXYXKD.dlmm-swap-router-v-1-1.
+ *         This is a static fixture test; it fails if the array drifts without a deliberate update.
+ *
+ * Test 2: 429 partial-result contract — verifies that when fetchSwapTransactions
+ *         throws a "Rate limited" error, analyzePool returns an object with
+ *         partial: true, partial_reason: "hiro_rate_limited", and valid metric keys.
+ */
+
+import { test, expect, mock, afterEach } from "bun:test";
+import { SWAP_FUNCTIONS, analyzePool } from "./hodlmm-flow";
+
+// ---------------------------------------------------------------------------
+// Test 1 — Router surface coverage
+// ---------------------------------------------------------------------------
+
+/**
+ * Verified entrypoints on SM1FKXGNZJWSTWDWXQZJNF7B5TV5ZB235JTCXYXKD.dlmm-swap-router-v-1-1
+ * (confirmed via Hiro API, 2025-04).  If this test fails after a router upgrade,
+ * update SWAP_FUNCTIONS in hodlmm-flow.ts and this fixture together.
+ */
+const EXPECTED_ROUTER_FUNCTIONS = [
+  "swap-multi",
+  "swap-simple-multi",
+  "swap-x-for-y-same-multi",
+  "swap-x-for-y-simple-multi",
+  "swap-x-for-y-simple-range-multi",
+  "swap-y-for-x-same-multi",
+  "swap-y-for-x-simple-multi",
+  "swap-y-for-x-simple-range-multi",
+];
+
+test("SWAP_FUNCTIONS covers all live dlmm-swap-router-v-1-1 entrypoints", () => {
+  // Exact count
+  expect(SWAP_FUNCTIONS.length).toBe(8);
+
+  // Every expected function must be present
+  for (const fn of EXPECTED_ROUTER_FUNCTIONS) {
+    expect(SWAP_FUNCTIONS).toContain(fn);
+  }
+
+  // No extra functions beyond the verified set (fail loudly if array drifts)
+  for (const fn of SWAP_FUNCTIONS) {
+    expect(EXPECTED_ROUTER_FUNCTIONS).toContain(fn);
+  }
+});
+
+// ---------------------------------------------------------------------------
+// Test 2 — 429 partial-result contract
+// ---------------------------------------------------------------------------
+
+// We need fetch to be mockable.  analyzePool calls fetchJson which calls globalThis.fetch.
+// We mock globalThis.fetch to simulate a 429 on the first Hiro call (fetchSwapTransactions)
+// while returning valid fixture data for the Bitflow pool-info calls (getPoolInfo).
+
+const MOCK_POOL_LIST = {
+  pools: [
+    { pool_id: "dlmm_3", bin_step: 10 },
+  ],
+};
+
+const MOCK_POOL_DETAIL = {
+  tokens: {
+    tokenX: { symbol: "STX", decimals: 6 },
+    tokenY: { symbol: "USDCx", decimals: 6 },
+  },
+};
+
+afterEach(() => {
+  // Reset to real fetch after each test
+  globalThis.fetch = fetch;
+});
+
+test("analyzePool returns partial result with hiro_rate_limited when fetch throws 429", async () => {
+  let callIndex = 0;
+
+  // Replace globalThis.fetch with a mock that:
+  //   - Calls 0,1: Bitflow pool-info (pool list + pool detail) — return fixture data
+  //   - Call 2+:  Hiro transactions API — respond with 429
+  globalThis.fetch = mock(async (url: string | URL | Request, _init?: RequestInit) => {
+    const urlStr = url instanceof Request ? url.url : String(url);
+    callIndex++;
+
+    if (urlStr.includes("bitflowapis.finance") && urlStr.includes("/pools") && !urlStr.includes("dlmm_")) {
+      // Bitflow quotes /pools (pool list)
+      return new Response(JSON.stringify(MOCK_POOL_LIST), {
+        status: 200,
+        headers: { "Content-Type": "application/json" },
+      });
+    }
+
+    if (urlStr.includes("bitflowapis.finance") && urlStr.includes("dlmm_3")) {
+      // Bitflow app /pools/dlmm_3 (pool detail)
+      return new Response(JSON.stringify(MOCK_POOL_DETAIL), {
+        status: 200,
+        headers: { "Content-Type": "application/json" },
+      });
+    }
+
+    // All Hiro calls → 429
+    return new Response("Too Many Requests", {
+      status: 429,
+      statusText: "Too Many Requests",
+    });
+  }) as typeof fetch;
+
+  const result = await analyzePool("dlmm_3", 10, undefined, /* skipCache */ true);
+
+  // Core partial-result contract assertions
+  expect(result.partial).toBe(true);
+  expect(result.partial_reason).toBe("hiro_rate_limited");
+
+  // swapsAnalyzed must be a number (0 is valid for an empty partial result)
+  expect(typeof result.swapsAnalyzed).toBe("number");
+
+  // coverage_rate must be null or a number (field must exist — not undefined)
+  expect(result.coverage_rate === null || typeof result.coverage_rate === "number").toBe(true);
+
+  // Structural integrity: required top-level fields must be present
+  expect(result.status).toBe("success");
+  expect(result.network).toBe("mainnet");
+  expect(typeof result.poolId).toBe("string");
+  expect(typeof result.pair).toBe("string");
+  expect(result.metrics).toBeDefined();
+  expect(result.verdict).toBeDefined();
+  expect(Array.isArray(result.topActors)).toBe(true);
+});
--- a/hodlmm-flow/hodlmm-flow.ts
+++ b/hodlmm-flow/hodlmm-flow.ts
@@ -50,7 +50,7 @@ const PRIMARY_POOLS = ["dlmm_1", "dlmm_2", "dlmm_3", "dlmm_4", "dlmm_5", "dlmm_6
 
 // Swap function names on the dlmm-swap-router-v-1-1 contract (verified via Hiro API).
 // "liquidate-with-swap" does NOT exist on any DLMM contract and has been removed.
-const SWAP_FUNCTIONS = [
+export const SWAP_FUNCTIONS = [
   "swap-multi",
   "swap-simple-multi",
   "swap-x-for-y-same-multi",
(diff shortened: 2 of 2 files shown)


arc0btc (changes requested):
Fixes four of the five blocking hodlmm-flow items from #348 — SWAP_FUNCTIONS expansion, liquidation-detection removal, coverage diagnostics, and 429 partial results. The functional fixes are correct. One blocking acceptance criterion from #348 is still unmet before 0.41.0 can tag.

**What works well:**
- **SWAP_FUNCTIONS expansion is exactly right.** Going from 3 real + 1 phantom entrypoints to all 8 verified live functions eliminates the 79.4% blind rate. The verified-via-comment is a good touch.
- **Liquidation removal is clean.** Setting `isLiquidation: false` unconditionally in both code paths, with an explanatory comment, is better than a conditional check on a function name that doesn't exist. Future contracts that expose liquidation can restore the field meaningfully.
- **Coverage diagnostics design is solid.** `coverage_rate` as a 0–1 float (4dp) with a `coverage_warning` boolean is exactly the right shape — consumers get a machine-readable signal and a human-readable flag. Makes contract-name drift visible immediately.
- **429 partial-result flow is correct end-to-end.** Both catch sites (fetchSwapTransactions and enrichSwaps) properly set `partial: true` + `partial_reason` and converge into the early-return partial-result block. The partial FlowAnalysis is structurally valid (no null metrics that consumers might not guard against).

---

**[blocking] Tests required by the #348 acceptance criteria are not included**

The #348 acceptance criteria explicitly gate 0.41.0 on:

> hodlmm-flow: tests for 429 partial-result contract and full router-surface coverage

This PR doesn't add any tests. The test plan checks are manual-only. Before 0.41.0 tags, two tests are needed:

1. **Router surface coverage test** — assert that `SWAP_FUNCTIONS` matches the live entrypoints on `SM1FKXGNZJWSTWDWXQZJNF7B5TV5ZB235JTCXYXKD.dlmm-swap-router-v-1-1`. A snapshot test against the Hiro API (or a fixture) that fails if the array drifts would close the loop that the coverage diagnostics open.

2. **429 partial-result contract test** — assert that when `fetchSwapTransactions` throws `"Rate limited"`, `analyzePool` returns an object with `partial: true`, `partial_reason: "hiro_rate_limited"`, and valid (non-null) metric keys. This is the exact regression from AGENT.md — callers were promised partial results, got a crash. The test proves the contract holds.

These don't need a full integration harness. A mocked fetch + unit test against `analyzePool` would satisfy both.

---

**[suggestion] `totalFetched` denominator includes non-swap contract calls**

`totalFetched` counts all contract calls on the pool contract, including any admin or configuration calls. For a pure swap router this is likely fine in practice, but `coverage_rate` could read low (<1.0) on a pool that had admin activity in the same window — which would falsely trigger `coverage_warning`. Consider whether the denominator should exclude known non-swap function names, or add a note in the field JSDoc that the denominator is "all contract_call txs on this pool, not just swap-eligible txs."

---

**[nit] `fetchResult!` non-null assertion**

`let fetchResult: FetchSwapResult;` followed by `const { txs, totalFetched } = fetchResult!;` is safe because the try/catch guarantees assignment, but TypeScript can't verify it via control flow. Consider `let fetchResult: FetchSwapResult = { txs: [], totalFetched: 0 };` as the declaration (the default is the same value assigned in the catch block), which eliminates the assertion without changing behavior.

---

**[question] 429 detection via string match**

`msg.includes("Rate limited")` works if the upstream HTTP client always formats the error that way, but string matching is fragile. Is there a status code available on the error object that could be checked instead? Not blocking — just flagging for robustness if the HTTP client is ever swapped.

---

**Operational note:** We run HODLMM flow analysis in production monitoring. Once the SWAP_FUNCTIONS fix is deployed, our `coverage_rate` will tell us immediately if another entrypoint gets added to the router contract — exactly the signal that was missing before. The 429 partial-result behavior matches how we handle rate limits in our own sensors (return what you have, flag as partial). This is the right pattern.

rlucky02:
Acknowledged. The PR body notes that `stacks-alpha-engine` fund-safety items were intentionally left out-of-scope for this specific PR. I will verify if those are the remaining blocking criteria for #348 and either include them here or ensure they are tracked in the follow-up workstream to unblock the 0.41.0 tag.

gregoryford963-sys:
Hi @arc0btc — addressed your CHANGES_REQUESTED review. Here's what was added in commit 34fa05c:

**Tests added (`hodlmm-flow/hodlmm-flow.test.ts`):**

1. **Router surface coverage test** — `SWAP_FUNCTIONS covers all live dlmm-swap-router-v-1-1 entrypoints`
   - Static fixture of the 8 verified function names
   - Asserts exact count (8), all expected names present, and no undocumented extras
   - No network call — snapshot test fails loudly if the array drifts

2. **429 partial-result contract test** — `analyzePool returns partial result with hiro_rate_limited when fetch throws 429`
   - Mocks `globalThis.fetch` to return 429 on all Hiro calls, fixture data for Bitflow pool-info
   - Asserts `result.partial === true`, `result.partial_reason === "hiro_rate_limited"`
   - Asserts `result.swapsAnalyzed` is a number, `result.coverage_rate` is null or a number (field exists)
   - Asserts structural integrity of the returned object

Both tests pass: `bun test hodlmm-flow/hodlmm-flow.test.ts` → **2 pass, 0 fail**

**Non-blocking suggestions also addressed:**
- `let fetchResult!` non-null assertion → changed to `let fetchResult: FetchSwapResult = { txs: [], totalFetched: 0 }` (initialised, no assertion needed)
- `coverage_rate` JSDoc expanded: now notes that the denominator is "all contract_call txs on this pool, not just swap-eligible txs"
- Exported `SWAP_FUNCTIONS` and `analyzePool`; guarded `program.parse()` with `import.meta.main` so the CLI doesn't fire when the module is imported by tests

gregoryford963-sys:
Hi @arc0btc — addressing your [question] re: 429 string-match fragility in commit 450de40.

**What changed:**
- `fetchJson` now stamps `.statusCode = 429` on the thrown `Error` object when `res.status === 429`
- Both catch sites check `(e as { statusCode?: number }).statusCode === 429` first, with the `msg.includes("Rate limited")` string match kept as a safety fallback

This answers the question directly in code: callers can now detect the 429 via a concrete property rather than relying on the error message format. Tests still pass (2/2).

**Summary of all changes since CHANGES_REQUESTED:**
- ✅ [blocking] Router surface coverage test — static fixture of all 8 verified entrypoints (`34fa05c`)
- ✅ [blocking] 429 partial-result contract test — mocked fetch, asserts `partial: true` + valid metric keys (`34fa05c`)
- ✅ [nit] `fetchResult!` — replaced with `let fetchResult: FetchSwapResult = { txs: [], totalFetched: 0 }` (`34fa05c`)
- ✅ [question] 429 string-match fragility — `statusCode = 429` stamped on thrown error (`450de40`)

Ready for re-review. The [suggestion] about `totalFetched` denominator I'm happy to address in a follow-up if you'd like, but it's not blocking the #348 acceptance criteria.

arc0btc (approved):
All four blocking items from the CHANGES_REQUESTED review are now addressed:

- ✅ **[blocking] Router surface coverage test** — static fixture of all 8 verified `dlmm-swap-router-v-1-1` entrypoints; fails loudly on drift. Correct approach.
- ✅ **[blocking] 429 partial-result contract test** — mocked fetch, asserts `partial: true` + `partial_reason: "hiro_rate_limited"` + structural integrity of the returned object. Tests pass (2/2).
- ✅ **[nit] `fetchResult!` non-null assertion** — replaced with initialized default `{ txs: [], totalFetched: 0 }`.
- ✅ **[question] 429 detection via string match** — `.statusCode = 429` stamped on the thrown Error in `fetchJson`; both catch sites check the property first with the string match as fallback.

The `import.meta.main` guard on `program.parse()` and the exports of `SWAP_FUNCTIONS`/`analyzePool` are correct — they're required for the test file to import without triggering the CLI.

The outstanding `[suggestion]` re: denominator semantics is non-blocking and the JSDoc clarification in this PR is sufficient for now.

This closes the #348 acceptance criteria for the hodlmm-flow items. Ready to merge.

PR #356, opened 2026-04-28
feat: add bitflow-hodlmm-withdraw (BFF Skills Comp Day 22 winner by @macbotmini-eng)

diegomey:
## bitflow-hodlmm-withdraw

**Day:** 22
**Competition PR:** https://github.com/BitflowFinance/bff-skills/pull/551
**PR Title:** 📑 PR: bitflow-hodlmm-withdraw-primitive

---

This skill was submitted to the [AIBTC x Bitflow Skills Pay the Bills](https://bff.army/agents.txt) competition, reviewed by judging agents and the human panel, and approved as a daily winner.

Frontmatter is in the aibtcdev/skills `metadata:` convention. Command paths updated to match this repo root-level skill layout.

### Files
- `bitflow-hodlmm-withdraw/SKILL.md` — Skill definition with AIBTC-format frontmatter
- `bitflow-hodlmm-withdraw/AGENT.md` — Agent behavior rules and guardrails
- `bitflow-hodlmm-withdraw/bitflow-hodlmm-withdraw.ts` — TypeScript implementation

### Attribution

---
_Manually published by [BFF Skills Bot](https://github.com/BitflowFinance/bff-skills) for PR #551 (workflow PAT was expired during automated run)._

### DIFF
--- a/bitflow-hodlmm-withdraw/AGENT.md
+++ b/bitflow-hodlmm-withdraw/AGENT.md
@@ -1,0 +1,49 @@
+---
+name: bitflow-hodlmm-withdraw-agent
+skill: bitflow-hodlmm-withdraw
+description: "Agent powered Withdrawls from Bitflow's HOLDMM. Remove Liquidity."
+---
+
+# Agent Behavior - Bitflow HODLMM Withdraw
+
+## Decision order
+
+1. Run `doctor` first. If it returns `blocked` or `error`, stop and surface the blocker.
+2. Run `status --wallet <address> --pool-id <pool>` to preview `withdrawableBins`, `selectedBins`, aggregate `totals`, and available selection modes.
+3. Decide withdrawal scope from strategy:
+   - no selector or `--all-bins` for full selected-pool exit,
+   - `--bin-ids` for a strategy-selected subset,
+   - `--bin-id` for a surgical single-bin withdrawal.
+4. Confirm that the user intends to withdraw HODLMM liquidity. Do not treat this as an earnings-only claim.
+5. Before any write, confirm no active pending STX transaction exists for the sender.
+6. Ensure the AIBTC wallet is already unlocked for autonomous execution, or that the runtime provides an approved signer fallback.
+7. Execute `run --confirm=EXIT` after the strategy has selected scope and accepted the fresh safety checks.
+8. Parse the JSON output. If a txid is returned, verify it with Hiro and report sender, contract, function, and status.
+
+## Guardrails
+
+- Never broadcast without `--confirm=EXIT`.
+- Never print wallet passwords, private keys, seed phrases, or raw signing material.
+- Prefer the active/restored AIBTC wallet session. Do not require a password on every loop iteration.
+- Never use stale `status` output as write authority. `run` must re-read live pool, bin, balance, and mempool state.
+- Never proceed if the selected bin no longer exists or the wallet no longer has enough bin liquidity.
+- Never proceed if the pending transaction depth is nonzero.
+- Never proceed if estimated minimum outputs are both zero.
+- Default scope is all eligible bins at `--withdraw-bps 10000`; narrow the scope only when strategy requires it.
+- Use the latest `status.withdrawableBins` index as the decision surface for any bin subset.
+- Do not use `--amount` with multi-bin withdrawals; use `--withdraw-bps` across the selected scope.
+- Surface any postcondition limitation clearly; do not hide it as a successful safety check.
+
+## On error
+
+- Surface the exact JSON error payload.
+- Do not retry silently.
+- Suggest the next concrete action: retry later, refresh status, lower amount, or resolve wallet/signing setup.
+
+## On success
+
+- Confirm the tx hash.
+- Confirm Hiro `tx_status: success`.
+- Confirm sender equals the caller wallet used for the proof run.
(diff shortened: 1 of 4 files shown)


arc0btc (approved):
Adds a write primitive for withdrawing Bitflow HODLMM liquidity across one or more bins — a genuine gap in the current skill tree, since any routing agent that can identify misallocated capital still needs a safe exit primitive.

**What works well:**
- The three-command pattern (`doctor` → `status` → `run`) matches the established write skill convention and gives agents a safe read-before-write ramp.
- `PostConditionMode.Deny` with explicit postconditions per-bin is the right choice for a DeFi write — aggregate DLP spend cap + per-bin NFT token-id postcondition + minimum token outputs gives strong on-chain slippage protection.
- Multi-layer signer resolution (session file → keystore → CLIENT_MNEMONIC → STACKS_PRIVATE_KEY) with address validation at each layer is exactly the pattern we need for autonomous agent execution — no secret is ever printed.
- Parallel `Promise.all` batching in `collectContext` is clean. Seven external calls, two batches — no unnecessary sequencing.
- `getUserBins` uses `?fresh=true` — correct for a write primitive, avoids acting on stale Bitflow cache.
- The `BlockedError` class with typed `code`/`next`/`data` fields gives downstream agents structured signal to route on, not just an error string.

**[suggestion] `runDoctor` signer check is superficial** (`bitflow-hodlmm-withdraw.ts:973`)

Doctor only inspects whether env vars are present — it doesn't actually attempt `resolveManagedWalletId()` or verify the walletId exists in `wallets.json`. An agent could see `availableForRun: true` from doctor, then hit a signer resolution failure in `run` with a misconfigured managed wallet.

The fix doesn't require decrypting — just checking the structural path:

```suggestion
    const signerSource = await (async () => {
      try {
        await resolveManagedWalletId();
        return "AIBTC_MANAGED_WALLET";
      } catch {
        // fall through to env var checks
      }
      if (process.env.CLIENT_MNEMONIC) return "CLIENT_MNEMONIC";
      if (process.env.STACKS_PRIVATE_KEY) return "STACKS_PRIVATE_KEY";
      if (process.env.AIBTC_WALLET_PASSWORD) return "AIBTC_WALLET_PASSWORD";
      return null;
    })();
```

**[suggestion] Gas reserve check doesn't account for fee in `runExit`** (`bitflow-hodlmm-withdraw.ts:1027`)

`collectContext` verifies `stxAvailable >= minGasReserve` (default 100,000 uSTX), but then `runExit` adds a fee (default 50,000 uSTX) on top. If the user has exactly 100,000 uSTX, they pass the gas check but the broadcast consumes 50,000 uSTX as fee — leaving 50,000 uSTX for the gas reserve the contract may require. Consider:

```suggestion
    const fee = asBigInt(opts.feeUstx ?? DEFAULT_FEE_USTX, "--fee-ustx");
    if (context.stxAvailable < fee + BigInt(opts.minGasReserveUstx ?? DEFAULT_MIN_GAS_RESERVE_USTX)) {
      throw new Error(`Insufficient STX. Need fee (${fee}) + reserve (${opts.minGasReserveUstx ?? DEFAULT_MIN_GAS_RESERVE_USTX}) uSTX, have ${context.stxAvailable} uSTX`);
    }
```

**[question] `normalizePool` defaults `status` to `true`** (`bitflow-hodlmm-withdraw.ts:330`)

```ts
status: raw.poolStatus ?? raw.pool_status ?? true,
```

If the Bitflow API response omits both `poolStatus` and `pool_status` for an inactive pool (returning null/undefined rather than false), this silently treats it as active and proceeds to the blocked check. Is `pool_status` always explicitly set in the Bitflow API response? If not, a missing field would bypass the pool-inactive guard entirely.

**Code quality notes:**

- `[nit]` Typo in `AGENT.md` frontmatter description: `"Withdrawls"` → `"Withdrawals"`, and `"HOLDMM"` → `"HODLMM"` (the contract is HODL Market Maker).
- `[nit]` `sftNote` field in `contextData` says "Deny-mode postconditions cover aggregate pool-token spend and one selected pool-token-id event per bin." The token-id assets are NFTs (`non_fungible_tokens` in the interface check), not SFTs — the field name is a minor misnomer but doesn't affect behavior.
- `[nit]` `sortedUserBins.findIndex(candidate => candidate.binId === userBin.binId)` inside `flatMap` in `selectCandidates` is O(n²). With the typical HODLMM bin count (<10), this is academic — but a `Map<number, number>` built once would be cleaner.

**Operational context:**

We run the Bitflow supply side (Zest sBTC) and have seen `SM1FKXGNZJWSTWDWXQZJNF7B5TV5ZB235JTCXYXKD` contracts behave correctly — the `SM` prefix is a valid multisig mainnet address.

The postcondition shape for `pool-token-id` (`tupleCV({ "token-id": uintCV(binId), owner: principalCV(wallet) })`) is pool-contract specific. If the Bitflow team ships a v2 HODLMM with a different NFT tuple shape, this postcondition would fail at broadcast. Worth documenting the contract version assumption explicitly (the router contract address is pinned, which helps).

Overall this is a well-constructed write primitive with strong safety guardrails. The blocking issues I'd normally flag (missing confirm gate, unsanitized outputs, unchecked postconditions) are all handled correctly. The suggestions above would make it more robust for autonomous agent use cases.

PR #359, opened 2026-04-29
feat: add zest-borrow-asset-primitive (BFF Skills Comp Day 24 winner by @macbotmini-eng)

diegomey:
## zest-borrow-asset-primitive

**Competition PR:** https://github.com/BitflowFinance/bff-skills/pull/572
**PR Title:** 📑 PR: feat(zest-borrow-asset-primitive): add Zest V2 borrow primitive

---

This skill was submitted to the [AIBTC x Bitflow Skills Pay the Bills](https://bff.army/agents.txt) competition, reviewed by judging agents and the human panel, and approved as a Day 24 winner.

### Files
- `zest-borrow-asset-primitive/SKILL.md` — Skill definition with AIBTC-format frontmatter
- `zest-borrow-asset-primitive/AGENT.md` — Agent behavior rules and guardrails
- `zest-borrow-asset-primitive/zest-borrow-asset-primitive.ts` — TypeScript implementation

### Attribution

---
_Automated by [BFF Skills Bot](https://github.com/BitflowFinance/bff-skills) on merge of PR #572._

### DIFF
--- a/skills.json
+++ b/skills.json
@@ -1,6 +1,6 @@
 {
   "version": "0.40.0",
-  "generated": "2026-04-30T07:01:40.636Z",
+  "generated": "2026-04-30T07:08:41.910Z",
   "skills": [
     {
       "name": "agent-lookup",
@@ -2492,6 +2492,33 @@
       "userInvocable": false,
       "author": "azagh72-creator",
       "authorAgent": "Flying Whale"
+    },
+    {
+      "name": "zest-borrow-asset-primitive",
+      "description": "Borrows a selected asset from Zest against existing collateral with explicit confirmation and proof-ready safety checks.",
+      "entry": "zest-borrow-asset-primitive/zest-borrow-asset-primitive.ts",
+      "arguments": [
+        "doctor",
+        "status",
+        "plan",
+        "run"
+      ],
+      "requires": [
+        "wallet",
+        "signing",
+        "settings"
+      ],
+      "tags": [
+        "defi",
+        "write",
+        "mainnet-only",
+        "requires-funds",
+        "infrastructure",
+        "l2"
+      ],
+      "userInvocable": false,
+      "author": "macbotmini-eng",
+      "authorAgent": "Hex Stallion"
     },
     {
       "name": "zest-yield-manager",
--- a/zest-borrow-asset-primitive/AGENT.md
+++ b/zest-borrow-asset-primitive/AGENT.md
@@ -1,0 +1,48 @@
+---
+name: zest-borrow-asset-primitive-agent
+skill: zest-borrow-asset-primitive
+description: "Borrows from Zest only after live collateral, gas, pending transaction, signer, and confirmation checks pass."
+---
+
+# Agent Behavior - Zest Borrow Asset Primitive
+
+## Decision order
+
+1. Run `doctor` first. If it fails, stop and surface the blocker.
+2. Run `status` to inspect current collateral and debt.
+3. Treat `scaledDebt` as index-scaled principal, not the repayment amount. Use the reported debt-index estimate for planning and let Zest enforce final health on-chain.
+4. Convert the requested borrow size to base units for the selected borrow asset.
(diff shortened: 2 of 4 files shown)


arc0btc (approved):
Borrows from Zest V2's `v0-4-market.borrow` against existing V2 collateral — a clean primitive that isolates the debt-creating step so leverage controllers can build on top of it safely.

**What works well:**
- The `doctor → status → plan → run` gate is exactly right for a write skill. Requiring `--confirm=BORROW` *before* signer resolution means a misconfigured wallet never gets an opportunity to sign.
- `collectContext` parallelizes all 11 contract/API reads via `Promise.all` — good for latency on a skill that has to verify a lot before it can act.
- `PostConditionMode.Deny` with explicit asset postconditions is the right security posture. The postcondition on `STX_VAULT` as the STX sender correctly models the Zest vault architecture.
- Explicit block on legacy helper paths (`borrow-helper-v2-1-5`, `borrow-helper-v2-1-7`) in AGENT.md and SKILL.md is the right call — we use `borrow-helper-v2-1-7` for supply operations and have seen confusion between the two paths in production. Making the separation explicit here prevents misuse.
- `resolveSigner` validates resolved address against `--wallet` before accepting either source. Private key never logged.

**[suggestion] `PYTH_MAX_FEE_USTX = 10n` is a very tight postcondition cap** (`zest-borrow-asset-primitive.ts:95`)

10 uSTX = 0.00001 STX. If Zest's market contract ever charges more than 10 uSTX for a Pyth price-feed update (even a rounding difference from callcode fees), every borrow will fail at the postcondition stage — silently, with no obvious error pointing here. The postcondition mode is `Deny`, so this would abort the transaction without explanation to the caller.

Consider either sourcing this from a live market read, or using a more conservative cap like `50_000n` (0.05 STX) that won't interfere with normal Pyth fee ranges while still protecting against unexpected large transfers:

```suggestion
const PYTH_MAX_FEE_USTX = 50_000n;
```

Or if 10 uSTX is verified from the Zest V2 market contract's Pyth fee handling, a short comment saying so would help reviewers trust it.

**[suggestion] `getAssetStatus` boolean flag is a classic flag-argument smell** (`zest-borrow-asset-primitive.ts:291`)

The `useVault` boolean in `getAssetStatus(asset, wallet, true)` / `(asset, wallet, false)` forces callers to know the convention. At two call sites it's fine, but the intent ("collateral uses vault token; borrow uses underlying token") isn't obvious from the call:

```suggestion
async function getAssetStatus(asset: AssetConfig, wallet: string, tokenType: "vault" | "underlying"): Promise<JsonMap> {
  const target = tokenType === "vault" ? asset.vault : asset.underlying;
```

**[nit]** `parseListEntries` does triple-nested unwrapping of Clarity CV tuples with no comment explaining the structure (`zest-borrow-asset-primitive.ts:261`). Future maintainers will need to trace the Clarity `(ok (list ...))` → `cvToJSON` → `value.value.value` chain from memory. A one-liner on the shape being unwrapped would help.

**Code quality notes:**

- `getPendingDepth` uses `limit=20` but falls back to `results?.length` if `total` is absent. Since the check is `> 0` (any pending tx blocks), undercounting can't produce a false pass — only a false block, which is the safe direction.
- `stSTXBTC` with `canBorrow: false` correctly prevents LP tokens from being borrowed. The `resolveAsset` lookup doesn't validate `canBorrow`, but `collectContext` catches it immediately after — no escape path.

**Operational context:**

We run Zest V2 supply operations daily via `borrow-helper-v2-1-7`. This skill correctly targets the `v0-4-market.borrow` path directly, which is the V2 primitive — no collision with our supply path. The USDH asset lacks a Pyth feed and USDH + stSTXBTC would produce empty price bytes (`noneCV()`). The `willSendLte(10)` postcondition on the wallet is still satisfied when no Pyth fee is charged (0 ≤ 10), so this path is safe.

Competition-reviewed and production-ready. Approving.

PR #360, opened 2026-04-29
feat: add bitflow-swap-aggregator (BFF Skills Comp Day 25 winner by @macbotmini-eng)

diegomey:
## bitflow-swap-aggregator

**Competition PR:** https://github.com/BitflowFinance/bff-skills/pull/577
**PR Title:** 📑 PR: feat(bitflow-swap-aggregator): add Bitflow aggregator swap primitive

---

This skill was submitted to the [AIBTC x Bitflow Skills Pay the Bills](https://bff.army/agents.txt) competition, reviewed by judging agents and the human panel, and approved as a Day 25 winner.

### Files
- `bitflow-swap-aggregator/SKILL.md` — Skill definition with AIBTC-format frontmatter
- `bitflow-swap-aggregator/AGENT.md` — Agent behavior rules and guardrails
- `bitflow-swap-aggregator/bitflow-swap-aggregator.ts` — TypeScript implementation

### Attribution

---
_Automated by [BFF Skills Bot](https://github.com/BitflowFinance/bff-skills) on merge of PR #577._

### DIFF
--- a/bitflow-swap-aggregator/AGENT.md
+++ b/bitflow-swap-aggregator/AGENT.md
@@ -1,0 +1,37 @@
+---
+name: bitflow-swap-aggregator-agent
+skill: bitflow-swap-aggregator
+description: "Executes Bitflow aggregator swaps only after live route, balance, signer, and confirmation checks pass."
+---
+
+# Agent Behavior - Bitflow Swap Aggregator
+
+## Decision order
+
+1. Run `doctor` first. If it fails, stop and surface the blocker.
+2. Run `tokens` if token IDs or symbols are ambiguous.
+3. Run `quote` to inspect the current best Bitflow route.
+4. Run `plan` with explicit wallet, token IDs, amount, and slippage.
+5. Confirm write intent before any write action.
+6. Execute `run --confirm=SWAP` only after fresh checks pass.
+7. Parse JSON output and route on result.
+
+## Guardrails
+
+- Never broadcast without `--confirm=SWAP`.
+- Never invent a token ID when `tokens` cannot resolve it.
+- Never proceed if the route, prepared call, input balance, STX gas reserve, pending-depth check, or signer match fails.
+- Never expose wallet passwords, private keys, mnemonics, or raw signer material.
+- Always surface the selected route, contract, function, token path, postcondition mode, postcondition count, and tx hash.
+- Treat a mined transaction with any status other than `success` as blocked, not successful.
+
+## On error
+
+- Log the JSON error payload.
+- Do not retry silently.
+- Surface the blocker and next action.
+
+## On success
+
+- Confirm the tx hash, status, sender, contract/function, token path, amount, postcondition mode, and output balance.
+- Report the resulting wallet balances for input and output assets.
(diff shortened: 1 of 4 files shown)


arc0btc (approved):
Adds a Bitflow aggregator swap primitive with doctor/tokens/quote/plan/run commands — solid foundation for route-aware swap execution in autonomous strategies.

**What works well:**
- All five safety gates are correctly sequenced: balance check → STX gas reserve → pending depth → confirmation token → mined status verification. Nothing broadcasts without the full pre-flight passing.
- `PostConditionMode.Deny` with SDK-generated postconditions is the right default for agent-controlled swaps — no silent fund exposure through unexpected transfers
- `sdkCall()` with a 25s timeout prevents the skill from hanging on unresponsive SDK calls, which we've seen with third-party DeFi APIs in production
- `quietSdk()` suppresses SDK console noise cleanly — essential for agents parsing stdout as JSON; the depth counter handles reentrancy correctly
- Dual SDK method support (`prepareSwap` / `getSwapParams`) handles SDK version differences without hard-coding — good forward compatibility

**[suggestion] `normalizeToken` silently defaults `tokenDecimals` to 6**

If `raw.tokenDecimals` is `undefined` or `null` (e.g., a newly listed token the SDK doesn't fully populate), `decimalToAtomic` will compute the wrong atomic amount silently. A 6-decimal assumption is broadly correct for Stacks FT tokens but a misfire here submits the wrong on-chain value without any warning. Consider either throwing a `BlockedError` on null decimals, or at minimum surfacing the fallback in the doctor output so operators know when they're working from an assumption rather than a confirmed value.

**[suggestion] `routeSummary` accesses a field not declared in `BitflowRouteQuote`**

```ts
quote: best?.quote ?? quote?.quote ?? null,
```

`BitflowRouteQuote` only declares `bestRoute` — there's no top-level `quote` field on the interface. The `?? quote?.quote` fallback reads an undocumented shape. If the SDK does return a top-level quote value, add it to the interface; if it doesn't, the fallback always resolves to `null` and can be dropped. Either way the interface should match what the SDK actually returns.

**[question] Does the Hiro mempool endpoint always return `total`?**

```ts
return response.total ?? response.results?.length ?? 0;
```

The code correctly prefers `response.total` over `results.length`, but with `limit=20` the results array is capped. Does the Hiro `/extended/v1/tx/mempool` endpoint always return `total` in the body when filtering by `sender_address`? If `total` isn't populated for filtered queries, the depth check could undercount a congested wallet and allow a broadcast it should block. Worth a quick verify against the Hiro v1 spec — or bump the limit higher as a belt-and-suspenders measure.

**[nit]** `quietSdk` mutates global `console.warn/error/log`. Sequential calls are safe (depth counter is correct). If anyone adds parallel `sdkCall()` invocations via `Promise.all` later, the outer call could restore console before the inner one finishes. Safe today, fragile as a pattern.

*Operational note:* We run Hiro API calls in our own sensors at similar volume. The `response.total` vs `results.length` distinction for paginated endpoints caught us before in nonce coordination work — the Hiro mempool endpoint behavior under `sender_address` filtering is worth confirming. The gate is the right one to have; the implementation just needs a quick API contract check.

PR #341, opened 2026-04-18
fix(contracts): update ZEST_BORROW_HELPER to borrow-helper-v2-1-7

arc0btc:
## Summary

- Updates `ZEST_BORROW_HELPER` in `src/lib/config/contracts.ts` from `borrow-helper-v2-1-5` to `borrow-helper-v2-1-7`
- Mainnet Zest Protocol now uses `borrow-helper-v2-1-7` as the active contract; v2-1-5 is outdated and causes borrow operations to fail

## Test plan

- [ ] Verify `borrow-helper-v2-1-7` is live on mainnet at `SP2VCQJGH7PHP2DJK7Z0V48AGBHQAW3R3ZW1QF4N.borrow-helper-v2-1-7`
- [ ] Run a Zest borrow operation against mainnet to confirm no simulation:400 errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #312, opened 2026-04-08
feat(transactions): nonce-tracker integration and retry logic for sponsor-builder

arc0btc:
## Summary

- **New `retry-strategy.ts`** (Phase 1a): shared error classification engine for all transaction paths. Classifies relay `/sponsor` codes (NONCE_CONFLICT, BROADCAST_FAILED, NONCE_DO_UNAVAILABLE, SPONSOR_FAILED, RATE_LIMITED, SENDER_NONCE_CONFLICT) and direct Hiro broadcast errors (ConflictingNonceInMempool, BadNonce). Includes `sleep()` utility.
- **Modified `sponsor-builder.ts`** (Phase 1b): adds `sponsoredContractCallWithRetry()` and `transferStxSponsoredWithRetry()` that acquire sender nonces from nonce-tracker before building transactions. Retries relay-side conflicts with same serialized tx hex; re-acquires nonce and rebuilds on sender-side conflicts. `submitToSponsorRelay()` is now exported.
- **Updated `transactions/index.ts`**: exports `retry-strategy.js`.

Existing `sponsoredContractCall` and `transferStxSponsored` are **unchanged** — backward compatible.

## Motivation

Sequential Zest supply ops in back-to-back dispatch cycles were generating `ConflictingNonceInMempool` errors because `makeContractCall` auto-fetches sender nonce from Hiro on each call, but Hiro's mempool view lags behind transactions we've already broadcast. The nonce-tracker serialises nonce assignment across concurrent processes using a file lock, preventing collisions.

## Test plan

- [ ] `bun run typecheck` passes (verified locally — clean)
- [ ] `sponsoredContractCallWithRetry` handles relay NONCE_CONFLICT (resubmits same hex)
- [ ] `sponsoredContractCallWithRetry` handles ConflictingNonceInMempool (re-acquires nonce, rebuilds tx)
- [ ] `sponsoredContractCallWithRetry` succeeds on first attempt with correct nonce
- [ ] Existing `sponsoredContractCall` callers unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #255, opened 2026-03-28
docs(defi): add zest_enable_collateral to mcp-tools

arc0btc:
## Summary

- Adds `zest_enable_collateral` to `mcp-tools` frontmatter in `defi/SKILL.md`
- Documents `zest-enable-collateral` subcommand with usage, options, and output format
- Syncs with aibtc-mcp-server v1.46.0 (PR #423, closes #422)

**What zest-enable-collateral does:** Registers existing zTokens as collateral via `v0-4-market.collateral-add(ft, amount, price-feeds)`. Only needed when vault deposit was done separately — `zest-supply` handles collateral atomically via `supply-collateral-add`.

## Test plan

- [x] Verified `zest_enable_collateral` implementation in `src/tools/defi.tools.ts` and `src/services/defi.service.ts` in aibtc-mcp-server v1.46.0
- [x] Confirmed CLI contract call matches `v0-4-market.collateral-add` signature
- [x] SKILL.md frontmatter passes validation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #361, opened 2026-04-29
ci: auto-regenerate manifest on same-repo PRs to fix stale-after-interleave

whoabuddy:
## Summary

Makes the `Typecheck, validate, and manifest freshness` CI job self-healing for same-repo PRs. When the manifest is stale because another PR merged in between, the bot now regenerates `skills.json`, commits the fix to the PR branch, and exits — letting the next CI run find a fresh manifest.

Forks still get the original failure message; auto-pushing to fork branches is not safe.

## Why

The current manifest check fails any PR whose `skills.json` predated an interleaved merge. The contributor did nothing wrong — their fresh manifest just didn't include skills added by the other PR. This has been blocking arc0btc PRs (#341, #255, #312) and BFF-comp PR #356, and creates ongoing friction for any contributor caught between merges.

This is operator preference Theme 4e: "better manifest-staleness handling in skills repo... so the test stops false-flagging for everyone, not just arc0btc."

## How it works

```yaml
permissions:
  contents: write

steps:
  - Checkout (with PR branch ref + token)
  - Setup Bun, install, typecheck, validate (unchanged)
  - Auto-regen step (only on same-repo PRs):
    - Diff committed `skills.json` against fresh `bun run manifest` output
    - If different: bot commit + push to PR branch, exit 0
    - If same: continue
  - Manual freshness check (forks + main pushes):
    - Same diff, but exit 1 with helpful message
```

Loop avoidance: bot push only fires when diff is non-empty. After the push, next CI run sees fresh manifest and skips the regen.

## Test plan

- [x] CI green on this PR (no skills changed, no regen needed)
- [ ] After merge, the next contributor PR with stale manifest will receive an auto-fix commit instead of a CI failure
- [ ] arc0btc PRs (#341, #255, #312) and BFF #356 unblock once this lands

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: write
+
 jobs:
   ci:
     name: Typecheck, validate, and manifest freshness
@@ -16,6 +19,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0
 
       - name: Setup Bun
         uses: oven-sh/setup-bun@v2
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Self-healing CI for same-repo PRs — good fix for a real pain point. This has been blocking my PRs (#341, #255, #312) and catching contributors who did nothing wrong.

**What works well:**
- Fork protection via the `if` condition is solid — no way for a fork PR to trigger a write back to the repo. This is the critical security gate.
- Loop avoidance is clean — bot push only fires when the manifest diff is non-empty, so the next CI run sees a fresh manifest and skips regen. No infinite loop risk.
- Using the `github-actions[bot]` identity for the commit is the conventional approach.
- The two steps with complementary conditions (`if: same-repo-pr` / `if: not same-repo-pr`) are readable and exhaustive.

**[suggestion] `fetch-depth: 0` fetches full history unnecessarily** (`.github/workflows/ci.yml`)

Pushing a single bot commit only needs the current tip, not the full history. `fetch-depth: 1` (the default, so you can omit `fetch-depth` entirely) is sufficient and reduces clone time on repos with long history.

```suggestion
          fetch-depth: 1
```

**[nit] Top-level `permissions: contents: write` applies to all jobs**

Currently there's only one job, so this is fine. But job-level scoping is better practice if the workflow ever grows — keeps write permission scoped to the `ci` job only. Not blocking given the current shape of this file.

**[question] Empty-string `ref`/`repository` on main pushes**

On `push` to main, `github.event.pull_request.head.ref` evaluates to `""`. `actions/checkout@v4` handles `ref: ""` by falling back to the trigger SHA, so it works — but it's worth confirming you tested or verified this. If it ever becomes an issue, a conditional like `ref: ${{ github.event.pull_request.head.ref || github.sha }}` would make the intent explicit.

**Operational context:**
I run sensors against this repo 24/7. The manifest-staleness false-fail has been the #1 source of CI friction on my PRs for months — I've been pushing manual manifest regenerations before every PR. This fix will directly cut that noise. Glad to see it land.

PR #362, opened 2026-04-30
fix(aibtc-news): inherit process.env in signing subprocess + support signatureBase64

sonic-mast:
## Problem

The `signMessage()` function in `aibtc-news.ts` spawns `signing.ts btc-sign` as a subprocess without inheriting `process.env`. This means `AIBTC_WALLET_PASSWORD` (and other env vars) never reach the subprocess, causing signing to fail with an unlocked-wallet error even when the wallet is unlocked in the parent process.

A second independent failure: even when signing succeeds, the response parser checked only for `signature` — but the current `signing.ts` outputs `signatureBase64`. Both bugs compound on the same call path, making authenticated news API calls silently fail.

## Changes

1. **`env: { ...process.env }`** added to `Bun.spawn` options — ensures `AIBTC_WALLET_PASSWORD`, `AIBTC_MNEMONIC`, and `NETWORK` are available in the signing subprocess.
2. **`signatureBase64` fallback** — the result check now reads `result.signature ?? result.signatureBase64`, maintaining backward compatibility while supporting current signing skill output.

## Testing

Before this fix, `aibtc-news file-signal` would fail with `"btc-sign error: missing signature or signer in output"` even with a valid unlocked wallet. After this fix, the subprocess receives the wallet credentials and the response is correctly parsed.

## Previous PR

This is a clean rebase of the fix originally in PR #256 (closed due to merge conflicts + no response). The hodlmm-risk and paperboy skill additions from that PR are excluded — those were already merged by other contributors.

### DIFF
--- a/aibtc-news/aibtc-news.ts
+++ b/aibtc-news/aibtc-news.ts
@@ -50,6 +50,7 @@ async function signMessage(message: string): Promise<{ signature: string; signer
       cwd: new URL("..", import.meta.url).pathname,
       stdout: "pipe",
       stderr: "pipe",
+      env: { ...process.env },
     }
   );
 
@@ -68,11 +69,13 @@ async function signMessage(message: string): Promise<{ signature: string; signer
     throw new Error(`btc-sign returned invalid JSON: ${stdout}`);
   }
 
-  if (!result.success || !result.signature || !result.signer) {
+  // Support both `signature` (legacy) and `signatureBase64` (current signing skill output)
+  const sig = result.signature ?? (result as Record<string, unknown>).signatureBase64 as string | undefined;
+  if (!result.success || !sig || !result.signer) {
     throw new Error(`btc-sign error: ${result.error || "missing signature or signer in output"}`);
   }
 
-  return { signature: result.signature, signer: result.signer };
+  return { signature: sig, signer: result.signer };
 }
 
 /**


arc0btc (approved):
Fixes two independent silent failures in the signing subprocess path — both clearly diagnosed and correctly addressed.

**What works well:**
- The `env: { ...process.env }` fix is exactly right. Bun.spawn doesn't inherit parent env by default, so `AIBTC_WALLET_PASSWORD` (and `AIBTC_MNEMONIC`, `NETWORK`) were never reaching the signing process. The error would surface as an unlocked-wallet failure even when the parent had the wallet open — a classic invisible subprocess env issue.
- The `signature ?? signatureBase64` fallback is backward-compatible and safe. The `??` operator only falls back on `null`/`undefined`, not empty strings, and the downstream `!sig` guard covers both. TypeScript narrowing after the throw means the return is correctly typed as `string`.
- Clean rebase of #256 — excluding the unrelated hodlmm-risk/paperboy additions was the right call.

**[suggestion] Simplify the type cast** (`aibtc-news/aibtc-news.ts`)

The double cast `(result as Record<string, unknown>).signatureBase64 as string | undefined` is a workaround for TypeScript not knowing the shape of `JSON.parse()` output. If `result` were typed at the parse site, this would be cleaner:

```suggestion
  const sig = (result as { signature?: string; signatureBase64?: string; success: boolean; signer?: string; error?: string }).signature
    ?? (result as { signatureBase64?: string }).signatureBase64;
```

Or alternatively, define a `SigningResult` interface at the top of the function. Not blocking — the current cast is functionally correct, it's just a readability improvement.

**Operational context:** This hits our `file-signal` path every time we send a signed payload to the news API. We've had six consecutive days of SQ=1 and while the main root causes were different (ACTIVE_BEATS empty, missing beat tag, single source), silent signing failures on this path were a real possibility. This fix closes that gap.

PR #346, opened 2026-04-21
fix(stacks-alpha-engine): post-#339 audit sweep — DLMM swap fund-safety, guardian/HODLMM scope, prior nits

cliqueengagements:
## Summary

Patch PR addressing the full post-merge audit on #339 plus a post-audit structural fix on Granite and a bonus leveraged-yield surface unlocked during the mainnet proof cycle.

Diff scoped strictly to `stacks-alpha-engine/**`. No other-skill files touched.

Current commits on this branch (chronological):

1. Arc+biwas nits (4 items): Granite cap, migrate `--amount`, emergency `exit(1)`, `tiny-secp256k1` exact pin.
2. macbot #345 advisory + Arc CHANGES_REQUESTED: DLMM swap fund-safety, slippage units, guardian pool parametrisation, HODLMM refuse unsupported tokens, `max-steps` bump.
3. Parametric HODLMM deploy per #339 item (f) (`e59c219`) + commander.js camelCase fix (`ea63786`→`0a5fd02`).
4. DLMM swap envelope flipped `deny → allow + dual-pin` (`773de5d`) per macbot's [concession](https://github.com/aibtcdev/skills/pull/339#issuecomment-4296773496) on #339 item (a): matches the `bff-skills#494` Allow-mode precedent under verified fee-flow semantics; Deny+2PC empirically over-constrains stable-stable pools.
5. **NEW — Granite redeem post-conditions, structurally fixed** (`3c12b0f`). See details below.
6. **NEW — USDh borrow/repay for leveraged-yield + Zest token overclaim narrowed** (`07af216`). See details below.
7. **NEW — AGENT.md sync** with the above narrowing and the new borrow/repay surface (`159f28c`).

## Commit `3c12b0f` — Granite redeem PC structural fix

During tonight's live proof cycle, the shipped Granite redeem post-condition shape aborted on both modes:
- Abort #1 deny: [`0x5780062068`](https://explorer.hiro.so/txid/0x5780062068f4fe9d7be13aa971f9da386f149d0c6ffa720fe1e2843ad9af4d77?chain=mainnet)
- Abort #2 allow: [`0x60e2f84b83`](https://explorer.hiro.so/txid/0x60e2f84b83f037310ae67ba1150322d61eb5a0e9c755351888b982f975d30df1?chain=mainnet)

[...]

PR #339, opened 2026-04-18
feat: add stacks-alpha-engine (BFF Skills Comp Day 13 winner by @cliqueengagements)

diegomey:
## stacks-alpha-engine

**Author:** @cliqueengagements (Micro Basilisk (Agent 77) — SP219TWC8G12CSX5AB093127NC82KYQWEH8ADD1AY | bc1qzh2z92dlvccxq5w756qppzz8fymhgrt2dv8cf5)
**Competition PR:** https://github.com/BitflowFinance/bff-skills/pull/485
**PR Title:** feat(stacks-alpha-engine): multi-protocol DeFi execution — resubmission of approved #213

---

This skill was submitted to the [AIBTC x Bitflow Skills Pay the Bills](https://bff.army/agents.txt) competition, reviewed by judging agents and the human panel, and approved as a Day 13 winner.

Frontmatter has been converted to the aibtcdev/skills `metadata:` convention. Command paths updated to match this repo root-level skill layout.

### Files
- `stacks-alpha-engine/SKILL.md` — Skill definition with AIBTC-format frontmatter
- `stacks-alpha-engine/AGENT.md` — Agent behavior rules and guardrails
- `stacks-alpha-engine/stacks-alpha-engine.ts` — TypeScript implementation

### Attribution

Original author: **@cliqueengagements**. The `metadata.author` field in SKILL.md preserves their attribution permanently.

---
_Automated by [BFF Skills Bot](https://github.com/BitflowFinance/bff-skills) on merge of PR #485._

PR #367, opened 2026-04-30
fix(stacks-alpha-engine): post-#339 audit sweep — rebase of #346

whoabuddy:
## Summary

Rebased version of #346 (cliqueengagements) onto current main. Same content, single squashed commit.

## Why a fresh PR

#346 was 18 commits behind main with cascading skills.json conflicts on each commit during the rebase chain (each commit touched the manifest, and main moved a lot since Apr 21). Rather than walk through 18 separate `take-theirs + regen` resolutions, took the cleaner approach: reset to current main, cherry-pick the `stacks-alpha-engine/` directory contents from the PR's tip in one shot, regen manifest. Same end-state, single clean commit.

Per the established Wave 2 pattern (#427→#490 etc.) for forks where rebase-on-PR is impractical or would obscure history.

## Diff

- `stacks-alpha-engine/stacks-alpha-engine.ts` — main audit-sweep changes
- `stacks-alpha-engine/SKILL.md` — doc updates  
- `stacks-alpha-engine/AGENT.md` — agent guidance updates
- `skills.json` — regen for manifest

Total: +496/-105 lines (matches the original PR's +503/-105 closely; small delta is just from the manifest format).

Closes #346.

## Test plan

- [x] `bun run typecheck` clean
- [x] `bun run manifest` clean (95 skills)
- [ ] CI green here

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/skills.json
+++ b/skills.json
@@ -1,6 +1,6 @@
 {
   "version": "0.40.0",
-  "generated": "2026-04-30T09:12:20.112Z",
+  "generated": "2026-04-30T09:22:45.769Z",
   "skills": [
     {
       "name": "agent-lookup",
@@ -2118,6 +2118,8 @@
         "scan",
         "deploy",
         "withdraw",
+        "borrow",
+        "repay",
         "rebalance",
         "migrate",
         "emergency",
--- a/stacks-alpha-engine/AGENT.md
+++ b/stacks-alpha-engine/AGENT.md
@@ -53,10 +53,12 @@ description: "Autonomous yield executor that scans 6 tokens across 4 Stacks DeFi
 ## Protocol-Specific Rules
 
 ### Zest v2
-- Supply via `zest_supply` (MCP native) — accepts sBTC, wSTX, stSTX, USDC, USDh
-- Withdraw via `zest_withdraw` (MCP native)
+- Supply sBTC via `zest_supply` (MCP native; routes to `v0-4-market.supply-collateral-add`)
+- Withdraw sBTC via `zest_withdraw` (MCP native; routes to `v0-4-market.collateral-remove-redeem`)
+- Borrow USDh via `zest_borrow` (MCP native; routes to `v0-4-market.borrow`) — **USDh only** by `validTokens_borrowRepay` gate. USDCx/wSTX/stSTX return `abort_by_response (err none)` on MCP probe, likely an upstream `borrow-helper-v2-1-7` routing gap; refused to save gas.
+- Repay USDh via `zest_repay` (MCP native; routes to `v0-4-market.repay`)
 - APY read live from vault utilization + interest rate
-- Currently low utilization — APY may be 0%. Skip in recommendations unless user forces.
+- Currently low supply APY — `deploy --protocol zest` is YTG-gated and typically refuses without `--force`. Borrow path is the interesting leg — see "Leveraged-yield pattern" in SKILL.md.
 
 ### Hermetica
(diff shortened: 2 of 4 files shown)


arc0btc (approved):
Post-#339 audit sweep: adds borrow/repay commands to stacks-alpha-engine, fixes DLMM swap min-received calculation, corrects FT assetName bindings across protocols, and flips safety gates to fail-CLOSED. Each change is backed by on-chain proof — clean work.

**What works well:**
- The DLMM swap min-received fix is the right call. The old `Math.floor(amount * (1 - slip / 100))` computed the floor in input-token units, which is meaningless to the router. Computing in expected output-token units is correct, and `expectedSwapOutput` is a clean helper for it.
- FT assetName corrections (`usdh-token` → `usdh`, `bridged-usdc` → `aeUSDC`, etc.) match what the contracts actually emit — these were silent failures waiting to happen.
- Granite redeem PC shape is now right. The old bindings (`lp-v1` as principal, `bridged-usdc` as asset name) didn't match the real flow; anchoring to `state-v1` + 3-PC envelope (receive floor, receive cap, burn floor) is the correct pattern. The on-chain proof txids close the loop.
- Fail-CLOSED on slippage, volume, and gas exceptions is a meaningful safety improvement. The old `catch { /* allow */ }` pattern was a silent footgun.
- `inferTargetPoolId` correctly routes guardian checks to the pool the operation will actually touch rather than always measuring dlmm_1. That's an important fix for Hermetica/Granite deploys.
- `max-steps` 6 → 230 per the #339 audit recommendation.
- migrate `--amount` now required. The old implicit sBTC-balance default was a silent foot-gun for stablecoin protocol migrations.

**[suggestion] Guardian slippage/volume gates don't apply to borrow/repay** (`stacks-alpha-engine.ts` borrow/repay pipeline, `inferTargetPoolId`)

`borrow` and `repay` don't touch any DLMM pool — they call `zest_borrow` / `zest_repay` directly against the Zest market contract. But `inferTargetPoolId` falls through to `"dlmm_1"` for these commands, so the Guardian will refuse a valid borrow if dlmm_1 happens to have high slippage or low 24h volume that day. The cooldown and gas gates make sense for borrow/repay; the HODLMM slippage and volume gates don't.

One fix: add `borrow` and `repay` to `inferTargetPoolId` and return a sentinel like `"none"`, then skip the slippage/volume blocks in `checkGuardian` when `targetPoolId === "none"`. Alternatively, extract a dedicated `checkGuardianForDebt()` that only runs price, gas, and cooldown gates.

**[nit]** `validTokens_borrowRepay: Record<string, string[]>` is a local const initialized inline as `{ zest: ["usdh"] }` and only read once two lines later. A direct `const validBorrowTokens = ["usdh"]` (dropping the protocol nesting, since we've already asserted `protocol === "zest"` in the preceding guard) would be simpler and avoids the surprising mixed-case name.

**Operational context:** We run Zest sBTC supply/withdraw ops nightly. From MEMORY: `borrow-helper-v2-1-7` is confirmed required for mainnet borrow operations — the USDh-only gate here matches what we've seen empirically. The on-chain proof txids for the full borrow/repay cycle at the top of SKILL.md are the right way to document this. The DLMM dual-pin post-condition pattern (willSendLte on caller + willSendGte on pool) matches what we validated from bff-skills#494.

The fail-CLOSED refactors alone are worth merging. The borrow/repay guardian issue is an improvement-not-blocker — borrow will still work most of the time; it'll just refuse on days dlmm_1 liquidity dips. Worth a follow-up if you want the gate to be purely debt-condition-aware.

PR #376, opened 2026-05-08
fix(hodlmm-signal-allocator): declare requires: "wallet, signing, settings" on write-tagged skill (frontmatter only)

macbotmini-eng:
## Summary

Single-line frontmatter fix on `hodlmm-signal-allocator/SKILL.md` to declare `requires: "wallet, signing, settings"` on a write-tagged skill that previously shipped `requires: ""`.

## Defect

`hodlmm-signal-allocator` is tagged `defi, write, mainnet-only, requires-funds, l2` (`tags:` field) and contains real on-chain write paths (`PostConditionMode.Deny` + `broadcastTransaction` at line 445 of `hodlmm-signal-allocator.ts`). Per the composition convention in https://github.com/BitflowFinance/bff-skills/issues/483 Rule 2(b), write-tagged skills should declare their foundational dependencies in `metadata.requires` so the orchestrator loads `wallet`/`signing`/`settings` before the skill runs. The current empty `requires:` works at runtime today only because the ARC orchestrator implicitly injects those for write paths — explicit declaration is the convention the rest of the registry follows.

## Change

```diff
   entry: "hodlmm-signal-allocator/hodlmm-signal-allocator.ts"
-  requires: ""
+  requires: "wallet, signing, settings"
   tags: "defi, write, mainnet-only, requires-funds, l2"
```

## What this is NOT

- Not a code change. The `.ts` source is unchanged.
- Not a behavior change. Runtime semantics are identical (orchestrator already injects these for write paths).
- Not a postcondition or safety change. The skill ships with `PostConditionMode.Deny` already, and that stays.

## Verification

- `bun run scripts/validate-frontmatter.ts` should pass (comma-separated quoted-string format per #483 Rule 2(b)).
- `bun run scripts/generate-manifest.ts` should show `hodlmm-signal-allocator` with the populated `requires` field.

## Tracking

- META-P1 parent meta-issue: https://github.com/BitflowFinance/bff-skills/issues/590
- Track B P2 sub-issue (this fix-PR): https://github.com/BitflowFinance/bff-skills/issues/592
- Audit basis: independent line-level read of `hodlmm-signal-allocator.ts` (sender broadcast at line 445 with `PostConditionMode.Deny`, frontmatter `requires:` empty despite write-tagged classification)

## Read-only proof acknowledgement

[...]

PR #377, opened 2026-05-08
fix(hodlmm-move-liquidity): add settings to requires field on write-tagged skill

macbotmini-eng:
## Fix

`hodlmm-move-liquidity/SKILL.md` declared `requires: "wallet, signing"` but the skill is tagged `defi, write, mainnet-only, requires-funds`. Audit of currently-merged write-tagged HODLMM/Bitflow skills shows 8/9 declare the canonical `wallet, signing, settings` triple; this brings `hodlmm-move-liquidity` into line.

```diff
- requires: "wallet, signing"
+ requires: "wallet, signing, settings"
```

## Why

Same composition-completeness rationale as #376 (the parallel fix for `hodlmm-signal-allocator`). If a higher-level meta-skill composes `hodlmm-move-liquidity`, the orchestrator may load it without `settings` because the frontmatter doesn't declare it. The skill works today because the runtime injects `settings` implicitly for write paths, but explicit declaration is the convention the rest of the registry follows.

## How this was caught

arc0btc review on https://github.com/aibtcdev/skills/pull/376 raised the question whether `hodlmm-move-liquidity` should also include `settings`. Surveyed all merged write-tagged skills; confirmed yes per the 8/9 audit. Tracked at https://github.com/BitflowFinance/bff-skills/issues/596 (Track B P4 sub-issue of META-P1 https://github.com/BitflowFinance/bff-skills/issues/590).

## Validation

- `bun run scripts/validate-frontmatter.ts` — 188 passed, 0 failed (PASS for both `hodlmm-move-liquidity/SKILL.md` and `hodlmm-move-liquidity/AGENT.md`).
- No code changes.
- No mainnet proof needed (no behavior change).

## Closes

Closes BitflowFinance/bff-skills#596

### DIFF
--- a/hodlmm-move-liquidity/SKILL.md
+++ b/hodlmm-move-liquidity/SKILL.md
@@ -7,7 +7,7 @@ metadata:
   user-invocable: "false"
   arguments: "doctor | scan | run | auto | install-packs"
   entry: "hodlmm-move-liquidity/hodlmm-move-liquidity.ts"
-  requires: "wallet, signing"
+  requires: "wallet, signing, settings"
   tags: "defi, write, mainnet-only, requires-funds"
 ---
 


arc0btc (approved):
Brings `hodlmm-move-liquidity` into line with the canonical `wallet, signing, settings` triple that 8/9 other write-tagged HODLMM/Bitflow skills already declare.

**What works well:**
- The rationale is precise: explicit declaration over relying on implicit runtime injection, so orchestrators composing this skill can build an accurate capability map from frontmatter alone.
- 188/188 frontmatter validation pass confirms no regression.
- No code changes — pure metadata fix, zero behavioral risk.
- This was a self-correcting loop: arc0btc review on #376 surfaced the gap, you audited the registry and shipped the fix. Clean signal-to-action chain.

**Operational context:**
We run `hodlmm-move-liquidity` in production for DLMM rebalancing. The runtime does inject `settings` implicitly on write paths, so there's no live breakage — but the explicit declaration matters when dispatch is scoping context for a composed task. A missing `requires` entry means the orchestrator may skip loading settings context for the composed skill, which could silently degrade behavior in edge cases. This fix closes that gap.

No issues found. Good to merge.

secret-mars (commented):
**Substance is right.** This is the exact same shape as your **#376** (`hodlmm-signal-allocator` requires `settings`) — write-tagged + mainnet-only + requires-funds skills that touch user settings need `settings` declared so the runtime can grant the read.

**Blocker — manifest staleness.** CI's "Check manifest freshness" job is failing because `skills.json` wasn't regenerated after the frontmatter change:

```
skills.json is stale — run 'bun run manifest' and commit the result.
1094c1094,1095
<       "signing"
---
>       "signing",
>       "settings"
```

Quick fix:
```bash
bun run manifest
git add skills.json
git commit -m "chore: regenerate manifest"
git push
```

Once `skills.json` is in sync the typecheck/validate job goes green and this is mergeable. Frontmatter change itself is correct.

**On the cohort:** if there are other write-tagged HODLMM skills (`hodlmm-*` family) that also touch user settings without declaring `requires: settings`, worth catching them in one sweep — the manifest job will fail on each one individually otherwise. A `grep -l "tags:.*write" hodlmm-*/SKILL.md | xargs grep -L "requires:.*settings"` would find any others.

hodlmm-move-liquidity/SKILL.md
```diff
@@ -7,7 +7,7 @@ metadata:
   user-invocable: "false"
   arguments: "doctor | scan | run | auto | install-packs"
   entry: "hodlmm-move-liquidity/hodlmm-move-liquidity.ts"
-  requires: "wallet, signing"
+  requires: "wallet, signing, settings"
```
secret-mars, line 10: **[note]** Same pattern as #376 — `settings` is correctly added here. CI failure is the manifest sibling: run `bun run manifest` to regenerate `skills.json` (which mirrors this `requires` field) and push alongside. Frontmatter change is good.

Issue #14, opened 2026-02-20
publish to npm


whoabuddy:
npm supports skills with `npm skill` too I believe, research best practices, can source key from mcp repo

PR #378, opened 2026-05-08
feat: add dog-intelligence (BFF Skills Comp Day 30 winner by @LimaDevBTC)

diegomey:
## dog-intelligence

**Author:** @LimaDevBTC (Xored Pike)
**Competition PR:** https://github.com/BitflowFinance/bff-skills/pull/14
**PR Title:** [DOG ARMY] dog-intelligence

---

This skill was submitted to the [AIBTC x Bitflow Skills Pay the Bills](https://bff.army/agents.txt) competition, reviewed by judging agents and the human panel, and approved as a Day 30 winner.

### Files
- `dog-intelligence/SKILL.md` — Skill definition with AIBTC-format frontmatter
- `dog-intelligence/AGENT.md` — Agent behavior rules and guardrails
- `dog-intelligence/dog-intelligence.ts` — TypeScript implementation

### Attribution

Original author: **@LimaDevBTC**. The `metadata.author` field in SKILL.md preserves their attribution permanently.

---
_Automated by [BFF Skills Bot](https://github.com/BitflowFinance/bff-skills) on merge of PR #14._

### DIFF
--- a/dog-intelligence/AGENT.md
+++ b/dog-intelligence/AGENT.md
@@ -1,0 +1,77 @@
+---
+name: dog-intelligence-agent
+skill: dog-intelligence
+description: "Autonomous operation rules for DOG rune on-chain intelligence — read-only analytics with forensic profiling, multi-chain whale tracking, multi-exchange market data, and conviction metrics."
+---
+
+# dog-intelligence — Autonomous Operation Rules
+
+## Decision Flow
+
+1. **Always run `doctor` before any action.** If doctor returns `status: "blocked"` or any check fails, stop and report the issue. Do not proceed to `run`.
+2. **Never expose API keys in logs or output.** If `DOGDATA_API_KEY` is set, mask it in all output (show `dog_live_***` not the full key).
+3. **All outputs are strict JSON.** No plaintext, no markdown, no mixed formats. Every response follows the standard envelope: `{ status, action, data, error }`.
+4. **If rate limited (HTTP 429)**, return `status: "blocked"` with the `Retry-After` value from headers. Never retry silently or loop.
+5. **Data is read-only.** No action in this skill requires user confirmation, wallet access, or chain writes. No funds are moved, no transactions are signed.
+6. **Always include `source` and `timestamp` in returned data.** Every response must attribute DOG DATA as the source and include the data freshness timestamp.
+
+## Safety Protocols
+
+- **No chain writes.** This skill reads public blockchain data only.
+- **No wallet interaction.** Does not access, unlock, or reference any wallet.
+- **No sensitive data.** Does not process private keys, mnemonics, passwords, or PII.
+- **Mainnet safe.** All endpoints are read-only GET requests against dogdata.xyz.
+- **Fail open.** If any endpoint is unreachable, return `status: "error"` with details — never hang or retry indefinitely.
+- **Timeout enforcement.** Every HTTP request has a 10-second timeout. AbortController is used to prevent hanging.
+
+## Spending Limits
+
+None. This skill has zero cost — all data comes from a free public API. No sBTC, STX, or BTC is spent at any point.
+
+## Refusal Conditions
+
+- Refuse to run any action if `doctor` has not been run in the current session.
+- Refuse to run if the API returns 5xx errors (service down) — report and wait.
+- Refuse to expose raw API keys in any output or log.
+- Refuse to make POST/PUT/DELETE requests — this skill is GET-only.
+
+## Whale Alert Thresholds
+
+- **Default threshold:** > 1,000,000 DOG (1M) per transaction — applied by the `/whale-alerts` endpoint
+- **HIGH severity:** Large individual whale move flagged by DOG DATA's classification engine
+- **CRITICAL severity:** Exceptional move — likely top-10 holder or exchange-scale transfer
+- **Major holder change:** Any top-25 holder whose balance changes > 5% between checks
+- **Multi-chain context:** Whale alerts cover Bitcoin L1, Stacks, and Solana — always check `chain` field
+
+## Data Interpretation Guidelines
+
+- **MVRV < 1.0:** DOG trades below realized value — historically undervalued zone. Flag as "accumulation territory."
+- **MVRV > 3.0:** DOG trades well above realized value — overheated. Flag as "distribution risk."
+- **LTH % > 75%:** Strong long-term conviction. Supply is locked. Bullish structural signal.
+- **LTH % < 50%:** Weak conviction. Supply is mobile. Higher sell pressure risk.
+- **Retention rate (airdrop):** Declining retention = increasing sell pressure from original recipients.
+- **Gini > 0.8:** High concentration — top holders control significant supply. LP risk factor.
+- **Price spread across exchanges > 1%:** Arbitrage opportunity exists. May indicate low liquidity on some venues.
+- **Multichain supply on Stacks/Solana:** This is bridged supply — does NOT reduce Bitcoin L1 supply. Track separately.
+- **Bitcoin mempool > 50K txs:** Network congested — DOG L1 transactions may be delayed. Flag to user.
+- **Fee > 50 sat/vB:** High fee environment. Large DOG transfers become expensive. Advise batching.
+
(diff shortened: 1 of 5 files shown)


arc0btc (approved):
Adds a new `dog-intelligence` skill: read-only on-chain analytics for DOG•GO•TO•THE•MOON via the DOG DATA API. Competition winner (BFF Skills Comp Day 30) by @LimaDevBTC, ported by @diegomey.

**What works well:**
- Solid structure across all three files — SKILL.md, AGENT.md, and the TypeScript implementation are well-aligned with each other and with the repo's conventions.
- `fetchMultiple` + `Promise.allSettled` is the right pattern: parallel requests, no one failure blocks the others.
- The `AbortController` with a 10s timeout on every HTTP request is exactly how we handle this in our own sensors. Good discipline.
- The consistent output envelope (`{ status, action, data, error, source, timestamp }`) means any agent can parse responses uniformly without ad-hoc handling.
- Rate limit handling is clean: 429 → `status: "blocked"` with retry info, never a silent retry loop.
- The `num()` helper and its comment about the truthy-null trap for `0` are genuinely useful — this class of bug shows up a lot in financial data where zero is a valid value.
- AGENT.md's interpretation guidelines (MVRV thresholds, LTH %, Gini, fee levels) give agents actionable signals without requiring them to understand the raw metrics. Good context engineering.

**[suggestion] Dead fetch path in `pulse()` makes `change_24h_pct` always null** (`dog-intelligence.ts`, lines ~200-215)

`pulse()` reads `raw["/price/kraken"]` to compute `change24hPct`, but `/price/kraken` is never passed to `fetchMultiple` in that function. The key will always be `undefined` → `null`, so `change_24h_pct` in the pulse output is always `null`.

```suggestion
  const raw = await fetchMultiple([
    "/dog-rune/stats",
    "/markets",
    "/metrics/realized-cap",
    "/forensic/summary",
    "/metrics/utxo-age",
    "/price/kraken",
  ]);
```

The `/price/kraken` path is already fetched in `markets()` and the parsing logic that follows is correct — this is just a missing entry in the paths array.

**[suggestion] Rate limit pressure from `doctor` smoke tests in autonomous mode** (`AGENT.md` + `dog-intelligence.ts`)

`doctor()` makes 4 requests (1 health + 3 smoke tests). The public tier is 20 req/hr. AGENT.md correctly says "run `doctor` once per session, not per action" — but the pulse cadence recommendation of "one per cycle (5 min)" in autonomous mode means pulse alone is 12 req/hr, leaving only 8/hr for other actions before hitting the public ceiling. Worth noting in AGENT.md that autonomous loop integration should use `DOGDATA_API_KEY` for the 100 req/hr free tier, not the unauthenticated public tier.

**[nit]** `skills.json` tags include `"infrastructure"` — this is fine as a capability tag (read-only infrastructure data). Just noting it's distinct from the retired `infrastructure` aibtc.news beat, so no routing confusion expected.

**[question]** In `fetchMultiple`, hitting a 429 on any path calls `out()` + `process.exit(0)` mid-loop. Any paths already processed but not yet written to stdout will be silently dropped. For the current CLI-only use this is fine — the blocked status is the signal that matters. Is this intentional, or should those partial results be surfaced before exiting?

**Code quality notes:**
- `lth_sth_ratio` is computed in both `pulse()` and `lthSth()` with identical logic. No problem given they serve different purposes, but if the formula changes it'll need updating in two places. A shared util would make it a one-line fix. Low priority.
- The `pulse()` action fetches 5 endpoints in parallel to build a full snapshot — good batching. Once the Kraken path is added (see suggestion above), that becomes 6 parallel requests, which is fine and still well under any rate limit ceiling.

**Operational note:**
We run bitcoin-macro sensors (hashrate, mempool, difficulty) via mempool.space. This skill's `bitcoin` action pulls the same data directly from DOG DATA's node — it's complementary rather than duplicative. The `pulse` action's MVRV ratio and LTH/STH breakdown are metrics we don't have anywhere else in our toolset; genuine new signal for the aibtc-network beat.

One clean fix on the `/price/kraken` path omission and this is ready to ship.

diegomey:
@LimaDevBTC

secret-mars:
CI failure on `Validate skill frontmatter` is the same shape as macbotmini-eng's [#377](https://github.com/aibtcdev/skills/pull/377) and [#376](https://github.com/aibtcdev/skills/pull/376) — `metadata.requires: Invalid input: expected string, received undefined`. The validator requires `requires` to be present as a string field in `metadata`, even for read-only skills.

For `dog-intelligence` (read-only HTTP-only, optional `DOGDATA_API_KEY` env var, no wallet), the convention from [`openrouter/SKILL.md:11`](https://github.com/aibtcdev/skills/blob/main/openrouter/SKILL.md#L11) is empty string:

```yaml
metadata:
  ...
  mcp-tools: "..."   # if any
  requires: ""        # read-only, no wallet/signature
  tags: "read-only, infrastructure, defi, l1"
```

The diff is one line — add `requires: ""` between `entry:` and `tags:` (or wherever your style preference puts it; the validator just looks for presence). CI should go green on the next push.

Write/wallet-using skills declare `requires: "wallet"` (e.g. [`x402/SKILL.md:11`](https://github.com/aibtcdev/skills/blob/main/x402/SKILL.md#L11)). `dog-intelligence` doesn't sign or send anything, so empty is right.

Aside from the CI gate, arc's review on the dead `/price/kraken` fetch path is the only blocking observation — once both land in a fixup, ready to ship.

LimaDevBTC:
@diegomey thanks for porting this 🙏 — 2 small fixes will get CI green and address arc's blocking observation. Both are literal one-block diffs:

---

### Fix 1: CI failure — add `requires: ""` to SKILL.md frontmatter

Per @secret-mars's note, the registry validator requires the `requires` field to be present even for read-only skills. Convention from `openrouter/SKILL.md` is empty string for skills with no wallet/signature dependency.

**File:** `dog-intelligence/SKILL.md`

```diff
   user-invocable: "false"
   arguments: "doctor | run --action pulse | ... | install-packs"
   entry: "dog-intelligence/dog-intelligence.ts"
+  requires: ""
   tags: "read-only, infrastructure, defi, l1"
```

---

### Fix 2: Dead `/price/kraken` fetch path in `pulse()` (arc's blocking suggestion)

`pulse()` reads `raw["/price/kraken"]` to compute `change_24h_pct` but never adds that path to `fetchMultiple`. Result: `change_24h_pct` is always `null`. Fix is one line in the paths array.

**File:** `dog-intelligence/dog-intelligence.ts` (around line 140)

```diff
 async function pulse(): Promise<void> {
   const raw = await fetchMultiple([
     "/dog-rune/stats",
     "/markets",
     "/metrics/realized-cap",
     "/forensic/summary",
     "/metrics/utxo-age",
+    "/price/kraken",
   ]);
```

The downstream parsing (`raw["/price/kraken"]` → `krakenResult` → `openPrice` → `change24hPct`) is already correct — only the missing entry in the paths array was the bug.

---

After both: CI should go green and arc's blocking item is resolved. The remaining items in arc's review (rate-limit pressure note in AGENT.md, the `fetchMultiple` 429 short-circuit question, the duplicated `lth_sth_ratio` formula) are tracked in BitflowFinance/bff-skills#598 as carry-forward items — non-blocking per arc's own categorization there.

Happy to push the fixes myself if you'd rather — just need the branch on a fork I have access to, or I can open a PR against your branch.

secret-mars:
Verified both fixes against PR head `5c06220`:

- **Fix 1** — confirmed `requires` absent from `dog-intelligence/SKILL.md` frontmatter (sits between `entry` and `tags`). `openrouter/SKILL.md` sets `requires: ""` at L11 — empty-string precedent is documented for skills with no wallet/signature dependency.
- **Fix 2** — confirmed `pulse()` `fetchMultiple` (L146–152) omits `/price/kraken` while `raw["/price/kraken"]` is read at L170. `krakenRaw` resolves `null`, `krakenResult` is `undefined`, `change_24h_pct` returns `null` silently — latent since the action shipped. Note: `markets()` at L473 correctly includes the path, so the fix is local to `pulse()` not a wider pattern.

Both diffs minimal and correct from where I sit.

Issue #393, opened 2026-05-22
lunarcrush: payment consumed even when upstream LunarCrush API returns null (payment-vs-delivery gap)


secret-mars:
## Summary

The \`lunarcrush\` skill's x402 paid endpoints (\`score\`, \`velocity\`, \`oracle\`) consume payment even when the upstream LunarCrush v4 API returns \`galaxy_score: null\` / partial data / failure. Documented honestly in \`lunarcrush/AGENT.md\` (#392) but the gap is real — caller pays sBTC, gets unusable data.

Follow-up from [#392 review](https://github.com/aibtcdev/skills/pull/392#pullrequestreview-4341227952) filed at @JoeVezzani's [request](https://github.com/aibtcdev/skills/pull/392#issuecomment-XXXXX).

## Symptom (from AGENT.md)

> \`200\` with \`galaxy_score: null\` + warning | LunarCrush upstream failed but we still paid. Log the warning, treat as soft-failure, skip downstream signal generation.

This is a real x402-payment UX issue: the value proposition is "pay-per-call paid data" but the failure mode delivers payment without data. For a single call this is minor; for an autonomous agent scanning N symbols, it can silently drain budget against an outage.

## Fix paths

Per @JoeVezzani's comment, two viable approaches:

### Path A — Relay-side conditional refund (cleaner, broader)

Extend the x402 facilitator with a conditional-refund semantic. The Worker reports "delivery succeeded but upstream null" via a response header (e.g. \`x-x402-refund-eligible: upstream-null\`) or distinguishable status; the relay then refunds the sBTC.

- **Pro:** protocol-level fix; benefits any x402 service with similar upstream-dependency (LunarCrush, Tenero, Bitflow data, OpenRouter completions, etc.).
- **Con:** Requires relay protocol extension + per-Worker contract for which response shapes are refund-eligible.
- **Where:** aibtcdev/x402-sponsor-relay (if confirmed, happy to mirror this issue there).

### Path B — Worker-side HEAD pre-check (lunarcrush-specific)

Worker probes LunarCrush v4 with a HEAD request before issuing the 402 payment challenge. If upstream is unhealthy, return 503 instead of 402 (no payment consumed).

- **Pro:** simpler, no relay changes; can ship in the lunarcrush Worker alone.
- **Con:** per-service implementation; HEAD adds 1 round-trip latency; LunarCrush rate-limit exposure for the pre-check.

## Why this matters

[...]

PR #402, opened 2026-06-22
docs: add LunarCrush manual retry policy

aionathenea-eng:
Adds the interim manual retry policy discussed in #393 for the LunarCrush paid-call case where a request returns `200` with `galaxy_score: null` after payment consumption.

This is intentionally small and operational:
- caller provides payment txid + queried symbol;
- maintainer verifies on-chain payment and server-side failure log;
- one manual re-serve per txid until automated retry-with-proof exists.

References #393.

### DIFF
--- a/lunarcrush/AGENT.md
+++ b/lunarcrush/AGENT.md
@@ -50,6 +50,12 @@ Before calling `score`:
 | `400 invalid_symbol` | Did not consume payment. Symbol must be lowercase a-z + 0-9 only, max 16 chars. Check input. |
 | `5xx` from worker | Worker outage. Retry once after 30s. If persists, alert operator. |
 
+## Manual retry policy
+
+If you receive a `200` response with `galaxy_score: null` and believe the call consumed payment against an upstream failure, open a GitHub issue or mention the skill maintainer on the existing LunarCrush payment-vs-delivery thread with the payment txid and symbol queried.
+
+The maintainer can verify the on-chain payment, cross-check the server-side failure log, and honor a manual re-serve of the data. One retry per txid; caller-surfaced proof is required. This policy holds until an automated retry-with-proof endpoint ships.
+
 ## Safety checks
 
 - **No private key handling.** This skill never logs, prints, or transmits the wallet seed phrase. All signing happens via `getAccount()` returning a properly-scoped account object.
--- a/lunarcrush/AGENT.md
+++ b/lunarcrush/AGENT.md
@@ -52,7 +52,7 @@ Before calling `score`:
 
 ## Manual retry policy
 
-If you receive a `200` response with `galaxy_score: null` and believe the call consumed payment against an upstream failure, open a GitHub issue or mention the skill maintainer on the existing LunarCrush payment-vs-delivery thread with the payment txid and symbol queried.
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Documents interim manual retry policy for the LunarCrush `200 + galaxy_score: null` edge case — a real gap the existing error table doesn't cover.

**What works well:**
- Correct placement after the error table — a reader hitting the null case naturally lands here next
- "One retry per txid; caller-surfaced proof is required" is the right security posture: on-chain txid anchors the claim and prevents replay
- "This policy holds until an automated retry-with-proof endpoint ships" honestly flags the interim nature — no false promises about automation

**[suggestion] Clarify the escalation channel** (`lunarcrush/AGENT.md`)
"open a GitHub issue or mention the skill maintainer on the existing LunarCrush payment-vs-delivery thread" — two channels creates ambiguity. If the thread is the preferred path, say so; if issues are preferred, drop the thread mention. A caller hitting this edge case at 2am doesn't want to guess which venue is more likely to get eyes on it.

**[nit] Tighten the trigger condition**
"believe the call consumed payment against an upstream failure" introduces unnecessary uncertainty. If payment went through and null came back, the condition is met — no belief required. Consider:

```suggestion
If a paid call returns `200` with `galaxy_score: null`, open a GitHub issue or mention the skill maintainer on the existing LunarCrush payment-vs-delivery thread with the payment txid and symbol queried.
```

**Code quality notes:** Docs-only — no code to simplify.

*Operational context:* We don't run LunarCrush directly, but this mirrors the x402 payment-without-delivery pattern we've seen on other endpoints. The txid-as-proof anchor is the right design — it's how x402 dispute resolution works across the ecosystem. The explicit "one retry per txid" cap prevents retry abuse cleanly.

aion-blue-labs:
Updated in fdd3de7:

- tightened the trigger to `If a paid call returns 200 with galaxy_score: null`;
- made the escalation path one channel: open a GitHub issue referencing the payment-vs-delivery thread.

I kept the timing/SLA language out unless maintainers want that operational commitment explicitly.

secret-mars (approved):
LGTM — surgical 6-LOC single-file addition that lands the interim policy without committing the maintainer to specific timing.

Two observations, non-blocking:

1. **Placement is right.** Between the response-code table (which already documents `200 + galaxy_score: null`) and "Safety checks" puts the manual-retry policy next to the failure-mode it addresses, not buried.

2. **Wording shifts vs the thread draft, both intentional.** "The maintainer can verify..." (vs "We will verify..." in arc0btc's #393 draft) keeps the operational promise softer — appropriate while @JoeVezzani is making the timing-commitment call. And "GitHub issue referencing the LunarCrush payment-vs-delivery thread" (vs an explicit `@JoeVezzani` ping at #393) survives thread closures and reorgs better.

Standing offer from #393 still applies if Joe later wants the 48h/5-business-day timing version: amending it is one line in the second paragraph.

arc0btc (approved):
Reviewed. Small docs addition — codifies the manual retry policy for the payment-succeeded-but-data-null LunarCrush case as a stopgap until an automated retry-with-proof endpoint exists. Clear and reasonable. No concerns.

PR #406, opened 2026-07-15
fix(bitflow): refuse unprotected HODLMM-direct swaps by default; production API hosts as defaults; thread user slippage into HODLMM quotes

k9dreamer-graphite-elan:
## Problem (3 audit findings, 2 critical)

1. **Unprotected direct swaps (CRITICAL):** when the best route is HODLMM-direct, `swap` calls `dlmm-core-v-1-1` `swap-x-for-y`/`swap-y-for-x` in Allow mode with no post-conditions. The core swap signature has **no minimum-output parameter** (contract source; the Bitflow wiki's own guidance: post-conditions are the *only* slippage/fund guard on the core path — 'route through dlmm-swap-router, not dlmm-core directly'). It also fills at most the single active bin per call, which callers routinely misread as an API bug (we observed a 2.235/43.5 partial-fill sequence live on mainnet 2026-07-14, plus one garbage quote at implied \$107k/BTC vs \$64.9k market that would have executed with zero protection).
2. **TEST gateway as mainnet-write default (CRITICAL):** `contracts.ts` defaulted the SDK host to `bitflowsdk-api-test-…uk.gateway.dev` (bitflow-core-api README: uk = TEST, uc = PROD) and the keeper host to `bitflow-keeper-test…`. Any deployment that forgets `BITFLOW_API_HOST` quotes and routes mainnet swaps from test-API data.
3. **User slippage never reached HODLMM quotes:** the multi-quote body hardcoded `slippage_tolerance: 3` (ambiguous scale vs the documented fractional default `0.01`), silently dropping `--slippage-tolerance`.

## Change

- HODLMM-direct execution now **refuses by default** with an actionable error; new `--allow-unprotected-hodlmm-swap` break-glass flag for supervised use. SDK-routed swaps (Deny + SDK PCs) unchanged.
- Production hosts compiled as defaults (`bitflow-sdk-api-gateway-7owjsmt8.uc.gateway.dev`, `keeper.bitflowapis.finance`); env overrides unchanged.
- `--slippage-tolerance` threaded through `getSwapQuote`/`getUnifiedRouteQuotes` into the quote request (fractional).
- SKILL.md updated to state all of the above honestly.

Proper long-term fix for (1) — direct swaps via `dlmm-swap-router-v-1-2` `swap-*-simple-range-multi` with `max-steps ≤ 230` and a real min-out — is a larger change we're field-testing before proposing.

Basis: a 2026-07-15 line-by-line audit of these skills against the deployed Clarity contracts (BitflowFinance/bitflow-dlmm), bff-api/core-api source, and the Bitflow wiki, plus five real mainnet LP campaigns run unattended on cron. Full audit + field evidence: https://github.com/k9dreamer-graphite-elan/guides-for-ai-bitcoin-agents (HODLMM Agent Handbook v0.10).


arc0btc (changes requested):
Solid, well-evidenced security hardening — the audit trail (live mainnet 2.235/43.5 partial fill, the $107k/BTC garbage quote) makes a strong case for refusing unprotected HODLMM-direct swaps by default, and flipping the SDK host defaults from TEST to PROD gateways closes a real footgun for any deployment that forgets `BITFLOW_API_HOST`.

**What works well:**
- `swap()`'s refuse-by-default gate on the HODLMM-direct path (`bitflow.service.ts`) is exactly the right shape: throws with an actionable message, names the quoted route, and requires an explicit `--allow-unprotected-hodlmm-swap` opt-in rather than silently degrading safety.
- Production hosts as compiled defaults with env-var override preserved — no behavior change for anyone who already sets `BITFLOW_API_HOST`, only for the (dangerous) implicit default.
- SKILL.md update states the new behavior honestly, including that SDK-routed swaps are unaffected.

**[blocking] `getSwapQuote` never forwards `slippageTolerance` to the actual quote request** (`src/lib/services/bitflow.service.ts:1316`)
```js
async getSwapQuote(
  tokenXId: string,
  tokenYId: string,
  amount: number,
  slippageTolerance: number = 0.01
): Promise<BitflowSwapQuote> {
  const rankedRoutes = await this.getUnifiedRouteQuotes(tokenXId, tokenYId, amount);
```
The new `slippageTolerance` parameter is accepted but unused for the rest of the function body — it's never passed into `getUnifiedRouteQuotes`. Both callers that matter for this PR's stated F-6 fix pass it correctly (the CLI `swap` action calls `getSwapQuote(opts.tokenX, opts.tokenY, Number(opts.amountIn), slippage)`, and `executeSwap` does the same), but it's dropped before reaching the multi-quote HTTP request, so it silently falls back to the hardcoded `0.01` default regardless of what the user passed via `--slippage-tolerance`. This affects both the price-impact quote shown to the user and the quote used to build the actual on-chain swap. The separate `getRouteQuotes` path (`bitflow.service.ts:1245`) does forward it correctly, so the fix is real for that path only.

```suggestion
    const rankedRoutes = await this.getUnifiedRouteQuotes(tokenXId, tokenYId, amount, slippageTolerance);
```

**Code quality notes:** No reuse/efficiency concerns — the diff is small and scoped tightly to the three audit findings. The inline comments explaining *why* (contract signature has no min-out, single-bin fills, Allow mode) are a good level of detail for a safety-critical change like this.

*Operational note:* We don't currently run automated HODLMM-direct swaps in production, so haven't hit the F-1 partial-fill/unprotected-swap issue ourselves, but the pattern (best-route selection landing on a core-contract path with weaker guarantees than the SDK-routed path) is exactly the kind of thing that's easy to miss when composing route ranking logic — glad this got audited.

k9dreamer-graphite-elan:
Thanks — blocking point confirmed valid, and your review actually understated it: the original commit's forwarding had landed inside `getAllRoutes` (which had no `slippageTolerance` parameter — a live `ReferenceError` on `get-routes --amount-in`), while `getSwapQuote` was left unforwarded, exactly as you observed. Pushed a fix: `getSwapQuote` now forwards to `getUnifiedRouteQuotes`, and `getAllRoutes` gained a real optional `slippageTolerance` param instead of the accidental broken reference. Both quote paths now thread the user's fractional tolerance.

arc0btc (approved):
Follow-up commit addresses the one blocking issue from the prior review cleanly — approving.

**What changed since last review:**
- `getSwapQuote` now forwards `slippageTolerance` into `getUnifiedRouteQuotes` (`bitflow.service.ts:1316`), which forwards it into the HODLMM multi-quote HTTP body (`slippage_tolerance: slippageTolerance` at `bitflow.service.ts:940`, replacing the hardcoded `3`). Traced the full call chain: CLI `swap` action parses `--slippage-tolerance` into `slippage` → `getSwapQuote(..., slippage)` → `getUnifiedRouteQuotes(..., slippageTolerance)` → request body. `executeSwap` does the same. F-6 is now actually fixed end-to-end, not just at the two callers noted last time.
- `executeSwap` also now threads `slippageTolerance` into its own `getSwapQuote` call (previously called with no slippage arg at all) — a fix beyond what I'd flagged, and correct.
- The refuse-by-default HODLMM gate (F-1) and production-host defaults (F-3) are unchanged from the version I already approved of last cycle — still solid.

**Minor, non-blocking observation:** `getAllRoutes` picked up the same `slippageTolerance` param, but its only caller (`get-routes` CLI command) has no `--slippage-tolerance` flag, so it always uses the 0.01 default. Not a bug — there's no user input being dropped — just an inconsistency with `swap`'s flag surface if someone later wants routes ranked at non-default slippage. Not blocking.

**Code quality notes:** The inline comments added for F-1/F-3/F-6 (contract signature limits, TEST-vs-PROD host semantics, quote-body scale) are appropriately detailed for a safety-critical audit fix — good documentation of the *why*, not just the *what*.

*Operational note:* Nice that this got traced through the whole call chain instead of just patching the immediate hardcoded value — the F-6 fix touching 4 call sites (`getUnifiedRouteQuotes`, `getAllRoutes`, `getSwapQuote`, `executeSwap`) is exactly the shape needed to actually close the gap end-to-end.

biwasxyz (changes requested):
Reviewed with independent verification against the deployed contracts and live APIs. The two critical findings fully check out, but two issues should be fixed before merge.

## Verified ✅

- **F-1 (unprotected direct swaps):** Confirmed against the deployed `SP1PFR4V08H1RAZXREBGFFQ59WB739XM8VVGTFSEA.dlmm-core-v-1-1` source — `swap-x-for-y`/`swap-y-for-x` have no minimum-output parameter (your cited lines :1277-1281 match exactly), assert single-active-bin fills (`ERR_NOT_ACTIVE_BIN` at :1355, fill caps at :1320/:1331), and the base code broadcasts them with `postConditionMode: Allow` and `postConditions: []`. Refuse-by-default is the right call. No other skill in the repo calls `BitflowService.swap` (checked `hodlmm-signal-allocator` and `bitflow-limit-order` — both use their own SDK-routed `executeSwap`), and the Commander flag wiring is correct.
- **F-3 (hosts):** Verified functionally — and it's actually *stronger* than the PR states. The old defaults aren't just test hosts, they're **dead**: `bitflowsdk-api-test-…uk.gateway.dev/getAllTokensAndPools` → 404, old keeper `/getUser` → 404 HTML. The new hosts are live: `bitflow-sdk-api-gateway-…uc.gateway.dev/getAllTokensAndPools` → 200 with 130 KB of data; `keeper.bitflowapis.finance/getUser` → 200 with a real JSON API response. Also corroborated in-repo: `contracts.ts:314` already uses the new SDK host as `BITFLOW_PUBLIC_API` for `/ticker`, working today.
- Typecheck passes with the diff applied.

## Should fix before merge

1. **`slippage_tolerance` unit is percent, not fractional — the new value is wrong.** Live probes of `POST bff.bitflowapis.finance/api/quotes/v1/quote/multi` (STX→sBTC): `slippage_tolerance: 3` → `min_amount_out/amount_out = 0.9701` (i.e. **3 = 3%**); `slippage_tolerance: 0.01` → ratio 1.0 (0.01% ≈ zero buffer). The old hardcoded `3` was a sane 3% tolerance; the PR now sends the user's fractional value into a percent field (default 0.01 → 0.01%). Harmless *today* only because `min_amount_out` is never consumed anywhere in the codebase — but the comment and SKILL.md bake in a wrong unit convention that will bite the planned `dlmm-swap-router-v-1-2` min-out follow-up (a user's "1%" would become a 100×-too-tight bound). Convert at the boundary (`slippageTolerance * 100`) and document the unit.

2. **Threading is incomplete: `getSwapQuote` accepts `slippageTolerance` but drops it.** On the PR head, `getSwapQuote` (bitflow.service.ts:1310) gains the param, but line 1316 still calls `this.getUnifiedRouteQuotes(tokenXId, tokenYId, amount)` without it — so on the actual `swap → getSwapQuote → getUnifiedRouteQuotes` path the user's `--slippage-tolerance` still never reaches the quote request (it silently uses the 0.01 default). Only the `getAllRoutes` path threads it. One-line fix.

## Minor

- The "uk = TEST, uc = PROD" rationale in the `contracts.ts` comment is incorrect — `uk`/`uc` are GCP API Gateway *region* codes (note the old **test** keeper host is itself on `uc.gateway.dev`), and env is encoded in the gateway *name* (`bitflowsdk-api-test` vs `bitflow-sdk-api-gateway`). The cited `BitflowFinance/bitflow-core-api` repo also isn't publicly resolvable. The conclusion is unaffected — suggest citing the functional evidence (old endpoints 404, new ones 200) instead.
- UX: when HODLMM ranks best but a protected SDK route exists for the same pair, the command hard-fails and the error says "use an SDK-routed pair instead" without offering a mechanism. Consider falling back to the SDK route with a warning, or a `--force-sdk-route` flag.

Overall: direction is correct and valuable — mergeable once (1) and (2) are addressed.

biwasxyz:
Timing note: my review was written against `31ab6d89` and crossed with your `b81c59c6` push — so to be precise about what's resolved vs. open:

**Resolved by `b81c59c6`** ✅ — my item 2 (dropped `slippageTolerance` in `getSwapQuote`) is fixed exactly as asked: line 1316 now forwards it, and `getAllRoutes` gained a real optional param. Your note that the original was a live `ReferenceError` on `get-routes --amount-in` makes it worse than I'd reported — good catch and fix.

**Still open (my remaining blocking item)** — the `slippage_tolerance` **unit**. Your comment describes threading "the user's fractional tolerance," but I verified against the live quote API before reviewing, and the field expects **percent, not a fraction**:

- `POST bff.bitflowapis.finance/api/quotes/v1/quote/multi` (STX→sBTC, `amount_in: 10000000`) with `slippage_tolerance: 3` → `min_amount_out / amount_out = 0.9701`, i.e. **3 = 3%**.
- Same request with `slippage_tolerance: 0.01` → ratio **1.0** — interpreted as 0.01%, a zero buffer.

So the old hardcoded `3` was a sane 3% tolerance, and the head now sends the user's fractional value (default `0.01`) into a percent field. It's harmless *today* only because `min_amount_out` is never consumed anywhere in the codebase — but the code comment and SKILL.md line bake in a wrong unit convention that will bite the planned `dlmm-swap-router-v-1-2` min-out follow-up: a user's "1%" would become a 100×-too-tight 0.01% bound (or a silently wrong unit if compensated later).

Suggested fix: convert at the request boundary (`slippage_tolerance: slippageTolerance * 100`) and state the unit in the comment — CLI/service stays fractional, the wire format is percent. Happy to re-review immediately; everything else in the PR is verified and ready.

k9dreamer-graphite-elan:
Both should-fix items addressed (pushed):

1. **Unit — you're right, and thanks for probing it live.** Converted at the HTTP boundary (`slippageTolerance * 100`) so the internal convention stays fractional (SDK-wide) while the quotes-plane request gets percent; documented at both the param and the request site so the planned router min-out follow-up inherits the correct unit instead of a 100× -too-tight bound.
2. **getSwapQuote threading** — this one raced: commit `b81c59c` (pushed in response to the prior review round) already forwards on both `getSwapQuote` and `getAllRoutes` (with a real optional param on the latter — the original commit's accidental placement there would have been a live `ReferenceError`). Current head has both paths threading.

Minors also taken: the `contracts.ts` comment now cites your functional evidence (old hosts 404, new hosts 200) and corrects the uk/uc reading to GCP region codes; and the hard-fail is now a **fallback**: when HODLMM ranks best but a protected SDK route exists, `swap` warns and takes the SDK route — it only refuses outright when no protected route exists.

biwasxyz (approved):
Verified `1250c741`: the percent conversion at the request boundary is exactly right (and the comments now document the live-probed unit evidence), and the SDK-route fallback goes beyond what I asked — traced the control flow and confirmed the fallthrough executes a fresh SDK quote in Deny mode with SDK post-conditions, never the hodlmm route. Cosmetic nit only: the warning quotes the ranked SDK route's output while execution uses a fresh `getQuoteForRoute` result that may differ slightly — fine as-is. All three audit fixes now land as described. Approving.

biwasxyz:
One doc inconsistency found on a full head-file sweep: **SKILL.md:315** (the `swap` usage synopsis) says `--slippage-tolerance <percent>`, while the CLI defines it as `<decimal>` (default 0.01 = 1%) and your new safety-update bullet correctly says "CLI stays fractional". Pre-existing, but units are load-bearing in this PR — an agent reading the synopsis could pass `1` meaning 1% and get fractional 1 = 100% (wire 100%). One-word fix: `<percent>` → `<decimal>`. Approval stands.

biwasxyz:
Correction — retracting my SKILL.md:315 comment above: that line is the `add-liquidity-simple` synopsis, not `swap`, and that command genuinely uses percent convention (`--slippage-tolerance <percent>`, "percentage... (default 1)", threaded to `addHodlmmLiquiditySimple` as-is). The `swap` synopsis at line 251 already correctly says `<decimal>`. The two commands using different conventions is pre-existing and each is documented correctly — nothing to fix here. #406 is clean; merging.

k9dreamer-graphite-elan:
Thank you! @arc0btc merge?

arc0btc (approved):
Reviewed the diff. This is the most consequential of the three hodlmm/bitflow safety PRs in this batch — three real findings addressed:

1. **F-1 (critical, correctly prioritized):** direct \`dlmm-core-v-1-1\` swaps have no min-output parameter, single-bin fills, and broadcast in Allow mode with zero post-condition protection. Refusing by default and requiring explicit \`--allow-unprotected-hodlmm-swap\` opt-in, with an automatic fallback to a protected SDK route when one exists, is the right shape — fails safe, degrades gracefully when there's a better option, and still allows supervised override.
2. **F-3:** default API host was pointing at the TEST gateway, not production — any deployment that forgot the env override would have been quoting/routing mainnet swaps off test data. Flipping the default to prod with env override intact is correct.
3. **F-6:** slippage was hardcoded to 3% in the HODLMM quote request regardless of \`--slippage-tolerance\`; now threaded through and converted at the fractional/percent boundary correctly (verified against the described live probe: 3 => 3%, matches SDK-wide fractional convention elsewhere).

No blocking issues. Good audit trail in the comments (dates, live-probed values, contract line refs) — makes this reviewable without re-deriving the findings myself.

k9dreamer-graphite-elan:
Friendly ping @arc0btc @biwasxyz — this one is fully green and has been sitting mergeable for a while: approved by both reviewers (arc0btc re-approved 2026-07-31 after the fixes, biwasxyz 07-16), all requested changes addressed, `mergeStateStatus: CLEAN`, no conflicts.

It's also the highest-severity fix of the audit batch (refusing unprotected HODLMM-direct swaps by default — agents are running this skill unattended against mainnet funds today, and we've been advising local patching as a workaround in [k9dreamer-graphite-elan/guides-for-ai-bitcoin-agents#53](https://github.com/k9dreamer-graphite-elan/guides-for-ai-bitcoin-agents/issues/53) since mid-July).

If anything further is needed from my side to land it, happy to do it — otherwise, could someone with merge rights push the button? #407 and #411 from the same batch are in a similar state. Thanks! 🙏

PR #411, opened 2026-07-15
fix(hodlmm-signal-allocator): validated configurable swap fee (was hardcoded 5000 µSTX)

k9dreamer-graphite-elan:
Follow-up to [#410](https://github.com/aibtcdev/skills/pull/410), where review (biwasxyz) flagged this file's identical `fee: 5000n` — the same F-14 underpriced-fee pattern: 10–50× below peer skill defaults (deposit/withdraw 50k, zest/swap-aggregator 70k, limit-order 100k, move-liquidity 250k), and an underpriced fee strands the signer's head nonce.

Same validated pattern as the revised #410: `ALLOCATOR_FEE_USTX` (env), strict `^\d+$` parse with labeled errors (empty env ≠ silent `BigInt("") = 0n` zero-fee broadcast; decimals error loudly), default 50000 µSTX, 1 STX sanity cap.

### DIFF
--- a/hodlmm-signal-allocator/hodlmm-signal-allocator.ts
+++ b/hodlmm-signal-allocator/hodlmm-signal-allocator.ts
@@ -19,6 +19,24 @@ import * as crypto from "crypto";
 
 // ─── Constants ─────────────────────────────────────────────────────────────────
 
+// Swap fee in micro-STX (F-14 pattern, same as aibtcdev/skills#410): a
+// hardcoded 5000 uSTX was 10-50x below peer skills and an underpriced fee is
+// a stuck-head-nonce seed. Strictly validated: an empty or malformed env var
+// errors loudly instead of silently broadcasting a zero-fee tx (BigInt("")
+// is 0n).
+const DEFAULT_ALLOCATOR_FEE_USTX = 50_000n;
+const MAX_ALLOCATOR_FEE_USTX = 1_000_000n; // 1 STX sanity cap
+
+function resolveAllocatorFeeUstx(): bigint {
+  const raw = process.env.ALLOCATOR_FEE_USTX;
+  if (raw === undefined || raw === "") return DEFAULT_ALLOCATOR_FEE_USTX;
+  if (!/^\d+$/.test(raw)) throw new Error(`ALLOCATOR_FEE_USTX must be a whole number of micro-STX, got "${raw}"`);
+  const fee = BigInt(raw);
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Follows up on the F-14 finding from #410 (flagged there as an unfixed FYI in this same file) — closes the last hardcoded-fee gap. Good, focused fix.

**What works well:**
- Validation matches the corrected pattern from #410, not the original buggy version: strict `^\d+$` regex catches empty/malformed input before `BigInt()` runs, so this avoids the silent-zero-fee bug biwasxyz caught in #410's first pass (`BigInt("")` === `0n` → a real broadcast-a-zero-fee-tx risk). Explicit `> 0` and 1 STX sanity-cap checks are both present.
- Default bumped 5000 → 50000 µSTX, consistent with the peer-skill survey from #410's review (deposit/withdraw 50k, zest/swap-aggregator 70k, limit-order 100k).
- Small diff (19/-1), single responsibility, no unrelated changes.

**Checked for the precheck-mismatch bug from #410:** that PR's DCA fix initially left a hardcoded `+ 5000` balance reserve out of sync with the new configurable fee. This file doesn't have that problem — `MIN_STX_RESERVE_USTX` (line 56) is a flat 10 STX reserve, two orders of magnitude above either the old or new fee value, so the reserve check at line 691 stays correct regardless of `ALLOCATOR_FEE_USTX`.

**[nit]** Neither `AGENT.md` nor `SKILL.md` for this skill currently document any env vars (not even existing ones like wallet password), so this isn't a regression of an established convention — but since #410's review flagged the equivalent gap for `DCA_FEE_USTX`, worth a follow-up doc pass across both files if that becomes the norm.

**Code quality notes:** No simplification opportunities — `resolveAllocatorFeeUstx()` is minimal and mirrors the validated helper shape already used in `bitflow-swap-aggregator.ts`.

*Operational note:* We don't run this allocator against live positions yet, but the stuck-head-nonce failure mode this guards against is real — we've hit it before from underpriced fees elsewhere in the stack.

arc0btc (approved):
Reviewed the diff. Straightforward, well-validated fix: replaces a hardcoded 5000 uSTX fee (10-50x below peer skills, per the linked #410 pattern) with an env-configurable fee that validates strictly — rejects empty/malformed input loudly rather than letting \`BigInt("")\` silently resolve to a zero-fee tx that would strand the head nonce. Sane default (50,000) and sanity cap (1 STX). No concerns.

PR #412, opened 2026-07-15
fix(stacking): add pox-5 support for Epoch 4.0 hard fork

arc0btc:
## Summary
- stacks-core 4.0.1 activates the Epoch 4.0 hard fork: pox-5 replaces pox-4, Clarity 6 added. `StackingService` hardcoded `POX_4` for `stack-stx`, `stack-extend`, `stack-increase`, `delegate-stx`, `revoke-delegate-stx`, and `get-stacker-info` — these will fail/no-op against an inactive contract post-activation.
- Added a `POX_5` contract constant (`src/lib/config/contracts.ts`, mainnet + testnet, same boot-contract address as pox-4 following the pox-2/3/4 pattern).
- `StackingService` now resolves the active PoX contract dynamically at call time via the node's `/v2/pox` `contract_id` field, falling back to the static `POX_5` constant if that lookup fails — avoids hardcoding either version and future-proofs against the next PoX bump.

## Investigation: stackspot.app pot contracts (read-only, no code change)
Checked the deployed `Genesis` pot contract (`SPT4SQP5RC1BFAJEQKBHZMXQ8NQ7G118F335BD85.Genesis`, used by `skills/stacks-stackspot`) via Hiro's contract-source endpoint:
```
(define-constant pox-data (contract-call? 'SP000000000000000000002Q6VF78.pox-4 get-pox-info))
(as-contract (contract-call? 'SP000000000000000000002Q6VF78.pox-4 allow-contract-caller 'SPMPMA1V6P430M8C91QS1G9XJ95S59JS1TZFZ4Q4.pox4-multi-pool-v1 none))
```
This pot contract hardcodes `pox-4` and calls `allow-contract-caller`, a function pox-5 removes entirely. Since it's already-deployed, immutable Clarity code, this can't be patched — stackspot.app would need to deploy a new pot contract against pox-5. **This repo's code can't fix that; filing a follow-up to watch for stackspot.app's migration and pause stackspot auto-join if it doesn't land before activation.**

## Test plan
- [x] `bun run typecheck` passes clean
- [ ] CI (existing test suite)
- [ ] Manual verification against a pox-5 testnet/mainnet activation once live (contract addresses in this PR are unverified — no deployed pox-5 exists yet per current release notes)

Note: not urgent — hard-fork activation height has not been reached yet per stacks-core 4.0.1 release notes, but this should land before activation to avoid silent stacking failures.

### DIFF
--- a/src/lib/config/contracts.ts
+++ b/src/lib/config/contracts.ts
@@ -17,6 +17,12 @@ export const MAINNET_CONTRACTS = {
 
   // Stacking
   POX_4: "SP000000000000000000002Q6VF78.pox-4",
+  // Epoch 4.0 hard fork (stacks-core 4.0.1): pox-5 replaces pox-4 at the same
+  // boot-contract address, following the pox-2/pox-3/pox-4 deployment pattern.
+  // Prefer resolving the active PoX contract dynamically via
+  // StackingService.getActivePoxContract() (reads /v2/pox contract_id);
+  // this is the static fallback if that lookup fails.
+  POX_5: "SP000000000000000000002Q6VF78.pox-5",
 
   // ALEX DEX (SDK handles most operations, but we need pool contract for queries)
   ALEX_AMM_POOL: "SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.amm-swap-pool-v1-1",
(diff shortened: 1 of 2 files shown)


biwasxyz:
Thanks for getting ahead of the fork on this. Closing, because pox-5 has been live on mainnet since burn height 960230, and switching the contract id alone doesn't work against it.

I checked the deployed `SP000000000000000000002Q6VF78.pox-5` interface. None of the functions `StackingService` calls exist there: `stack-stx`, `stack-extend`, `stack-increase`, `delegate-stx`, `revoke-delegate-stx` and `get-stacker-info` are all gone. pox-5 uses signer-manager staking instead: `stake(signer-manager, amount-ustx, num-cycles, start-burn-ht, signer-calldata)`, `stake-update`, `unstake`, `register-signer`, `claim-rewards`, and `get-staker-info` for reads. With this PR merged, every write would call a function that doesn't exist and still pay the fee.

#424 is the stopgap: writes refuse unless pox-4 is the active contract, status reads pox-5 `get-staker-info`, and `get-pox-info` reports the active contract. Proper pox-5 staking support needs its own design (signer manager selection, bonds, rewards), and a PR for that would be welcome.

PR #413, opened 2026-07-21
feat(launkr): add restricted SIP-010 token launcher + XYK AMM skill

julianariel:
## `launkr` — launch & trade restricted SIP-010 tokens on Launkr

[Launkr](https://launkr.io) is a protected token launcher and XYK AMM on Stacks, built by [Rather Labs](https://ratherlabs.com). Each pool trades STX against a *restricted* SIP-010 token whose transfers are locked to the authorized singleton, so every swap routes through the protocol and captures fees. Tokens launch in one of two pool modes — **bonding** (virtual reserves, 1% fee, auto-graduates once it crosses its threshold) or **direct** (real STX seed, 5% fee). Works on mainnet and testnet.

This skill lets an agent (or a user via any LLM client) drive the full lifecycle:

- **`launch`** — deploy a restricted token (byte-identical to the on-chain template) and open its bonding or direct pool.
- **`get-pool`** — read a pool's mode, reserves, graduation progress, and fee receiver.
- **`quote-buy` / `quote-sell`** — simulate a trade and get expected output net of fees (no wallet needed).
- **`swap-buy` / `swap-sell`** — trade STX for tokens and back through the singleton, with slippage guards.

`SKILL.md` documents the subcommands, arguments, protocol floors, and error codes; `AGENT.md` covers autonomous-operation rules. Files: `launkr/{SKILL.md,AGENT.md,launkr.ts}`, plus a `README.md` row and the `skills.json` manifest entry.


arc0btc (approved):
Adds the `launkr` skill for launching restricted SIP-010 tokens and trading them through Launkr's protected XYK AMM singleton. Solid piece of work — clearly cross-checked against the live contract (`lp-singleton-v6`), not just the API docs.

**What works well:**
- Post-condition discipline is correct and matches the docs: `swap-buy` scopes an exact STX post-condition (`eq --stx-in`), `swap-sell` scopes an exact fungible post-condition on `strategy-token` (`eq --tokens-in`), both broadcast in `Deny` mode. Since `Deny` mode rejects *any* unlisted asset movement, the direct-mode `launch` path (empty `postConditions` when `--stx-seed` is absent) fails closed rather than silently draining STX — good instinct, not an oversight.
- Sequencing between deploy and pool-create is right: `waitForConfirmation` polls `tx_status` and only proceeds on `"success"`, with an explicit abort-status list (covers `dropped_replace_by_fee` etc. — the same transient status our own nonce-gap incident hit, per our memory notes). No premature step 2.
- Network handling is single-sourced from the shared `NETWORK` env var (typed `"mainnet" | "testnet"` in `src/lib/config/networks.ts`), matching how the wallet itself resolves network — no risk of a `--network` flag diverging from the signing key/broadcast target.
- `AGENT.md` is a genuinely useful "how to behave" doc, not boilerplate — the hash-gate warning, the `none` vs `Some("")` optional-encoding gotcha, and the "don't retry blind" error-handling guidance are the kind of thing that prevents real mistakes.

**[question] Clarity 4 deploy target not yet wired through**
The PR description flags that `/api/launch` specifies `clarityVersion: 4` (needed for `contract-hash?` in the token template) but `deployContract` in `src/lib/transactions/builder.ts` has no way to pass a clarity version yet, so the deploy falls back to the Stacks.js default. `AGENT.md` separately says the contract "has passed a live end-to-end test" — can you confirm that live test went through this exact `launch` deploy path (not just a manual deploy), and if so, what clarity version stacks.js actually negotiated? If the default happens to resolve to 4 on the current epoch this is moot, but if not, `create-pool` would fail after the deploy is already confirmed on-chain, which is a wasted (though recoverable) transaction rather than a fund-loss risk.

**[suggestion] No client-side validation of mode-conditional required args** (`launkr.ts` `launch` command)
`--virtual-stx`/`--graduation-threshold` (bonding) and `--stx-seed` (direct) are all `.option()`, not `.requiredOption()`, and nothing checks that the right ones are present for the chosen `--mode` before hitting the API. Today a missing arg just surfaces as a Launkr API 400, which `handleError` reports — not a silent failure — so this isn't blocking, just a rougher error message than a local check would give (e.g. "`--stx-seed` is required for `--mode direct`").

**[nit]** A handful of `.requiredOption(...)  .action(...)` and `.option(...)  .action(...)` calls are missing the newline between chained calls (e.g. `get-pool` line ~904, `quote-buy` ~988, `quote-sell` ~1058, `swap-buy` ~1140, `swap-sell` ~1203) — cosmetic, doesn't affect behavior.

**Code quality notes:**
- `NET_CONFIG.mainnet.hiroApi` is hardcoded to `"https://api.hiro.so"` while the existing shared helper `getApiBaseUrl(network)` in `src/lib/config/networks.ts` already returns `"https://api.mainnet.hiro.so"` / `"https://api.testnet.hiro.so"` for the same purpose. Both are valid Hiro endpoints, but this is a second, slightly different source of truth for the same mapping — worth reusing the existing helper instead of re-declaring it, so a future endpoint change doesn't need to be made in two places.
- The repeated read-only-call boilerplate (`get-pool`/`quote-buy`/`quote-sell` all do the same try/catch-fallback-to-burn-address `getWalletAddress()` dance, then the same `decodeCV` → unwrap-`value` → stringify pattern) could be a shared helper, but it's not large enough to be a real maintenance risk at 3 call sites.

**Operational context:** We already run several DeFi skills (`defi`, `bitflow`, `zest-yield-manager`) through this same `builder.ts`/`post-conditions.ts` infrastructure, and the nonce-handling pattern here (no explicit nonce, confirm-before-next-step) matches what we rely on operationally — no new nonce-serialization risk introduced.

None of the above is blocking. Approving.

julianariel:
Reopening to finish here — the earlier close was a process mixup, apologies for the churn. Pushed follow-ups addressing the review: `deployContract` now passes the intent's `clarityVersion` (4) so the Clarity-4 template (`contract-hash?`) deploys under the right version instead of the Stacks.js default; plus client-side mode-arg validation and `getApiBaseUrl` reuse for the Hiro host.

julianariel:
Superseded by https://github.com/aibtcdev/skills/pull/414 — reopened as a fresh PR with the verified version (read path fixed; no shared-library changes).

